PR #89, opened 2023-05-15
feature: Connection state

kaladivo:


apps/mobile/src/state/marketplace/atom.ts
```diff
@@ -45,6 +49,14 @@ export const myActiveOffersAtom = focusAtom(myOffersAtom, (optic) =>
   optic.filter((myOffer) => myOffer.offerInfo.publicPart.active)
 )
 
+export function createMineOfferAtom(
```
SamTremko: pls remove not used

SamTremko (approved):

PR #94, opened 2023-05-17
Filter offers

SamTremko:
Offers for user filtered + my offers filter


apps/mobile/src/components/Dropdown/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+  size,
+}: Props<T>): JSX.Element {
+  const tokens = getTokens()
+  const [isOpen, setIsOpen] = useState<boolean>(false)
+  const activeRow = useMemo(
+    () => rows.find((row) => row.type === activeRowType),
+    [activeRowType, rows]
+  )
+
+  const rotation = useSharedValue(isOpen ? 180 : 0)
```
PanJarda, line 95: I suggest not to use magic strings in the code
```suggestion
  const rotation = useSharedValue(isOpen ? Angle.OPEN : Angle.CLOSED)
```

kaladivo: generally, I am for using constants instead of hard coded values. But in this case it's much more readable to have values directly instead of defined constants. What is open/closed angle? 

But nice catch. Thanks :)

apps/mobile/src/components/FilterOffersScreen/FilterOffersContent.tsx
```diff
@@ -0,0 +1,63 @@
+import {ScrollView, StyleSheet} from 'react-native'
+import ScreenTitle from '../ScreenTitle'
+import IconButton from '../IconButton'
+import refreshSvg from '../images/refreshSvg'
+import closeSvg from '../images/closeSvg'
+import OfferForm from '../OfferForm'
```
kaladivo: Filter form

apps/mobile/src/components/FilterOffersScreen/components/FriendLevel/images/secondDegreeFriendsSvg.ts
kaladivo, line 1: Move to componentsImages

apps/mobile/src/components/OfferForm/components/AmountOfTransaction/images/dashSvg.ts
kaladivo, line 1: Move up to prevent double definition.

kaladivo (approved):

kaladivo (approved):

PR #118, opened 2023-05-23
Logout improvements

SamTremko:
Offers are deleted in logout callback
Offer is removed also from offer-to-connections atom
Solved unhandled promise rejection on chats screen

### DIFF
--- a/apps/mobile/src/components/FilterOffersScreen/index.tsx
+++ b/apps/mobile/src/components/FilterOffersScreen/index.tsx
@@ -3,6 +3,7 @@ import {type RootStackScreenProps} from '../../navigationTypes'
 import React from 'react'
 import FilterOffersContent from './FilterOffersContent'
 import FilterOffersScopeProvider from '../FilterOffersScopeProvider'
+import useSafeGoBack from '../../utils/useSafeGoBack'
 
 type Props = RootStackScreenProps<'FilterOffers'>
 
@@ -10,16 +11,12 @@ function FilterOffersScreen({
   route: {
     params: {type},
   },
-  navigation,
 }: Props): JSX.Element {
+  const safeGoBack = useSafeGoBack()
   return (
     <Screen customHorizontalPadding={0} customVerticalPadding={32}>
       <FilterOffersScopeProvider type={type}>
-        <FilterOffersContent
-          navigateBack={() => {
-            navigation.goBack()
-          }}
-        />
+        <FilterOffersContent navigateBack={safeGoBack} />
       </FilterOffersScopeProvider>
     </Screen>
   )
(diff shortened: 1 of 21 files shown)


apps/mobile/src/components/FilterOffersScreen/index.tsx
```diff
@@ -3,23 +3,20 @@ import {type RootStackScreenProps} from '../../navigationTypes'
...
+  const safeGoBack = useSafeGoBack()
   return (
     <Screen customHorizontalPadding={0} customVerticalPadding={32}>
       <FilterOffersScopeProvider type={type}>
-        <FilterOffersContent
-          navigateBack={() => {
-            navigation.goBack()
-          }}
-        />
+        <FilterOffersContent navigateBack={safeGoBack} />
```
kaladivo, line 19: Great. It's better to use this. In the future we might open some screen directly by deepLink / notification and it might be problem calling goBack without checking if there is previous navigation state in the stack.

apps/mobile/src/components/ModifyOffer/atoms/offerFormStateAtoms.ts
```diff
@@ -264,16 +266,23 @@ export const offerFormMolecule = molecule((getMolecule, getScope) => {
...
         TE.match(
           (e) => {
             Alert.alert(
-              toCommonErrorMessage(e, t) ?? t('offerForm.errorCreatingOffer')
+              toCommonErrorMessage(e, t) ?? t('editOffer.errorDeletingOffer')
             )
             return false
           },
           (result) => {
+            set(offerToConnectionsAtom, (previousValue) => ({
```
kaladivo: How about moving this to deleteOffersAtom set action? It might be better idea since with every offer deletion we also want to delete the offer from offer to connections. What do you think?

apps/mobile/src/state/chat/hooks/useFetchNewMessages.ts
```diff
@@ -195,7 +195,7 @@ export default function useFetchMessagesForAllInboxes(): () => T.Task<'done'> {
             )()
         ),
         // @ts-expect-error bad typings?
```
kaladivo, line 197: If there is not error here, can you please remove //@ts-expect-error ? If there still is an TS error, can you remove this change from changelog? We'll look at this in another MR

apps/mobile/src/state/useHandleLogout.ts
kaladivo, line 1: Please rename the file to match the function name.

kaladivo (changes requested):

kaladivo (approved):

PR #141, opened 2023-05-28
Empty marketplace design

SamTremko:


kaladivo (changes requested):
We need to add better explanation of the reach number. Le't say user is the first in Argentina who has Vexl. We need to communicate that he needs to get others to install Vexl as well. 

We also need to communicate that the social number is not count of his imported numbers but it is count of how many **users of vexl** can he reach. 

We need to properly communicate following: 
- Copy of explanation of social network number.
- Show the reach number
- Tell user to get more of his contacts to use Vexl


If we do not communicate this, users will become frustrated about why they don't see any offers.

kaladivo (approved):

PR #142, opened 2023-05-28
feature: #80 - add force update and maintenance mode

### LINKED ISSUE #80: Force update & Maintenance mode
AC: 
- By changing the values in remote config app can be put into
  - Maintenance mode
  - Force update mode
- Should be separate configs for prod / stage
- Should be separate configs for old / next app

kaladivo:


apps/mobile/app.config.ts
```diff
@@ -6,6 +6,7 @@ const presets = {
   stage: {
     enableHiddenFeatures: true,
     apiPreset: 'stageEnv',
+    VERSION_CODE,
```
SamTremko: pls add versionCode

SamTremko (approved):

PR #146, opened 2023-05-29
feature: #90 - Add version to headers

kaladivo:
& move location api definition to src/api


SamTremko (changes requested):
Use versionCode instead of version ENV var

SamTremko (approved):

PR #167, opened 2023-05-30
fix: UI fixes for translated elements

SamTremko:


apps/mobile/src/components/InsideRouter/components/MarketplaceScreen/index.tsx
```diff
@@ -38,13 +41,12 @@ function MarketplaceScreen(): JSX.Element {
 
   const tabBarLabelStyle: StyleProp<ViewStyle> = useMemo(
     () => ({
-      fontSize: 40,
-      lineHeight: 42, // if we
+      fontSize: width < SCREEN_SIZE_THRESHOLD ? 30 : 40,
```
kaladivo: Would it be possible to use [mediaQueries](https://tamagui.dev/docs/core/use-media) here?

apps/mobile/src/state/chat/atoms/selectOtherSideDataAtom.ts
```diff
@@ -10,6 +10,7 @@ export default function selectOtherSideDataAtom(
   chatAtom: Atom<Chat>
 ): Atom<UserNameAndAvatar> {
   return selectAtom(chatAtom, (chat) => {
+    console.log(`Chat: ${JSON.stringify(chat, null, 2)}`)
```
kaladivo: This is probably leftover from debugging. Remove or change log to be more descriptive

apps/mobile/src/utils/localization/i18n.ts
```diff
@@ -1,13 +1,13 @@
...
+import cs from '../../localization/cs'
 import en from '../../localization/en'
-// import sk from '../../localization/sk'
+import sk from '../../localization/sk'
 
 const i18n = new I18n({
   en,
-  // cs,
-  // sk,
+  cs,
```
kaladivo: If you want to include this. Which makes sense. Please add condition on `enableHiddenFeatures`.

kaladivo (changes requested):

kaladivo (approved):
Please squash before merging

PR #171, opened 2023-05-31
bugfix/not-possible-to-change-profile-picture-on-onboarding-once-selected

SamTremko:


apps/mobile/src/components/LoginFlow/components/PhotoScreen/index.tsx
```diff
@@ -85,14 +90,14 @@ function PhotoScreen({
...
     ])
-  }, [t, reportAndTranslateErrors])
+  }, [t, reportAndTranslateErrors, forceToRerender])
 
   return (
     <>
       <HeaderProxy showBackButton progressNumber={1} />
       <WhiteContainer>
         <Stack maw="70%">
-          <Text numberOfLines={2} adjustsFontSizeToFit ff="$heading" fos={24}>
```
kaladivo, line 95: Why remove this? Is this breaking anything?

kaladivo (commented):
I think better approach is to actually change the image URI. Let's not trust system it will change the image if the same URI is provided to the ImageView (even if we change the key property). It depends on implementation, it might be cached somewhere...

apps/mobile/src/components/LoginFlow/components/PhotoScreen/index.tsx
SamTremko: This was causing, that the title on that screen was really small compared to design.

kaladivo (approved):

PR #172, opened 2023-05-31
bugfix: imported contacts count in profile section

SamTremko:


apps/mobile/src/components/InsideRouter/components/SettingsScreen/components/ButtonsSection.tsx
```diff
@@ -36,11 +36,36 @@ import openUrl from '../../../../../utils/openUrl'
...
 const ItemText = styled(Text, {
   fos: 18,
 })
 
+const SubtitleText = styled(Text, {
+  fos: 12,
+  col: '$greyOnBlack',
+})
+
+function ContactsImportedTitle({
```
kaladivo: Would it be possible to move these components to another file? This file is already super long. It might improve code readability.

kaladivo (approved):

Issue #198, opened 2023-06-05
Set time limit to encrypting offers background task so offers are updated at least partially


kaladivo:
On iOS there is a slow crypto, if the app receives update to re encrypt offers for lot of users, it will fail, make sure we encrypt it at least partially.

PR #199, opened 2023-06-05
feature: #198 - Set time limit to encrypting offers background task s…

kaladivo:
…o offers are updated at least partially


apps/mobile/src/state/connections/atom/offerToConnectionsAtom.ts
```diff
@@ -112,14 +151,30 @@ export const updateAllOffersConnectionsActionAtom = atom(
...
                 reportError(
                   'error',
                   'Error while encrypting new connections for offer',
                   encryptionErrors
                 )
               }
+
+              if (timeLimitReachedErrors.length > 0) {
+                reportError(
+                  'error',
```
SamTremko: change to warn

SamTremko (approved):

Issue #234, opened 2023-06-09
Add more fiat currencies


prusnak:
CAD, HUF, PLN, JPY, etc.

prusnak:
First step in this direction: https://github.com/vexl-it/vexl/pull/373

leomira:
also seems fine!

PR #336, opened 2023-06-19
Simplify currency handling

### LINKED ISSUE #335: Use currency codes consistently
Please make the use of currency shortcuts consistent across the app.

Currently, there is "Czech crown" in the menu, "CZK" in create offer and then suddenly "Kč" appears in the marketplace.

I don't have *strong* preference if we should use currency codes "CZK / EUR / USD" or symbols "Kč / $ / €", but it should be at least consistent in the settings menu, create offer form and marketplace.

prusnak:
Please review carefully since:
1. this is my first bigger PR to this repo
2. this code is completely untested

This PR:
- simplifies currency handling - i.e. removes unnecessary translations and transformations of used currency = just use CZK/USD/EUR everywhere
- simplifies implementing https://github.com/vexl-it/vexl/issues/234 for the time being
- fixes https://github.com/vexl-it/vexl/issues/335

We can later reintroduce something like the mapping:

```
USD => $
CZK => Kč
EUR => €
```

but not via translations, but rather via a fixed map (dict) in the code.

This creates a new problem, though, where for example both CNY and JPY use the same symbol ￥ and THB uses the symbol ฿. So we might never introduce this.

### DIFF
--- a/apps/mobile/src/components/OfferInfoPreview.tsx
+++ b/apps/mobile/src/components/OfferInfoPreview.tsx
@@ -52,19 +52,8 @@ function OfferInfoPreview({
   )
 
   const offerAmount = useMemo(() => {
-    if (offer.publicPart.currency === 'CZK')
-      return `${bigNumberToString(offer.publicPart.amountTopLimit)} ${t(
-        'common.czkSymbol'
-      )}`
-    if (offer.publicPart.currency === 'EUR')
-      return `${t('common.eurSymbol')}${bigNumberToString(
-        offer.publicPart.amountTopLimit
-      )}`
-    if (offer.publicPart.currency === 'USD')
-      return `${t('common.dollarSymbol')}${bigNumberToString(
-        offer.publicPart.amountTopLimit
-      )}`
-  }, [offer.publicPart.amountTopLimit, offer.publicPart.currency, t])
+    return `${bigNumberToString(offer.publicPart.amountTopLimit)} ${offer.publicPart.currency}`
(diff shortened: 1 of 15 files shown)


apps/mobile/src/components/OfferInfoPreview.tsx
```diff
@@ -52,19 +52,8 @@ function OfferInfoPreview({
...
-    if (offer.publicPart.currency === 'EUR')
-      return `${t('common.eurSymbol')}${bigNumberToString(
-        offer.publicPart.amountTopLimit
-      )}`
-    if (offer.publicPart.currency === 'USD')
-      return `${t('common.dollarSymbol')}${bigNumberToString(
-        offer.publicPart.amountTopLimit
-      )}`
-  }, [offer.publicPart.amountTopLimit, offer.publicPart.currency, t])
+    return `${bigNumberToString(offer.publicPart.amountTopLimit)} ${offer.publicPart.currency}`
```
kaladivo, line 55: Since the app is not displaying currency symbols in this code change, it is ok always display currency after the amount, right?

kaladivo (commented):
Have you run `yarn utils:sync-translations`? 

Also, if you are removing translations from the translations files. Make sure to remove them from the table as well. Currently, the script does only modify table for new translations. It does not removes missing translations from the table. This is to prevent accidental deletion of translations.

apps/mobile/src/components/OfferInfoPreview.tsx
prusnak: Yes, that was my idea for this change.

kaladivo (approved):

PR #340, opened 2023-06-21
feature: first degree offers should be always visible regardless of c…

### LINKED ISSUE #341: Social network updates & Explanation
As discussed at #291  I would like to continue a discussion regarding the updated approach towards "direct friend" situation in the social network. 

Below I propose a quick fix solution (before we rewrite the contact microservice) with a UX suggestion. 

We have the information popup already designed (attached here), I would reuse it in the request dialogue.

I suggest this wording:
"What does this mean?" + button "Check it out" or "Learn more"

On the popup that follows we should communicate:

situation no1 - you are a direct friend with 0 common friends
headline: Offer from a direct friend
You see this offer because the counterparty has your phone number saved in their contact list. Be cautious, we cannot verify if you really know each other in the real life.

situation no2 - you are a direct friend with XY common friends
headline: Offer from a direct friend
You see this offer because the counterparty has your phone number saved in their contact list. Don't forget to verify their identity with a common contact.

situation no3 - you are a 2nd degree friend with XY common friends
This is a contact with whom you have no direct connection. Don't forget to verify their identity with a common contact.

I think it would be best to always show the popup according to the state of the request rather than show one long dialogue explaining all three.

Snímka obrazovky 2023-06-21 o 11 17 15

### LINKED ISSUE #358: First degree offers should be always visible regardless of common friends count

kaladivo:
…ommon friends

### DIFF
--- a/apps/mobile/src/components/Info.tsx
+++ b/apps/mobile/src/components/Info.tsx
@@ -1,0 +1,54 @@
+import {getTokens, Stack, Text, XStack} from 'tamagui'
+import SvgImage from './Image'
+import infoSvg from './images/infoSvg'
+import {TouchableOpacity} from 'react-native'
+import closeSvg from './images/closeSvg'
+import Button from './Button'
+import {useState} from 'react'
+
+interface Props {
+  actionButtonText: string
+  text: string
+  onActionPress: () => void
+}
+
+function Info({
+  actionButtonText,
+  text,
+  onActionPress,
+}: Props): JSX.Element | null {
+  const tokens = getTokens()
+  const [isVisible, setIsVisible] = useState<boolean>(true)
+
+  if (!isVisible) return null
+
+  return (
+    <Stack jc={'center'} p={'$4'} bc={'$pinkAccent1'} br={'$4'}>
+      <XStack ai={'center'} justifyContent={'space-between'} mb={'$4'}>
+        <XStack f={1} space={'$2'} ai={'center'}>
+          <SvgImage fill={tokens.color.pink.val} source={infoSvg} />
+          <Stack fs={1}>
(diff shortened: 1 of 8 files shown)


prusnak:
Are we sure that UI does not break for offers with no common friends?

SamTremko (approved):

apps/mobile/src/components/OfferDetailScreen/components/OfferInfo.tsx
```diff
@@ -17,8 +21,60 @@ import InfoSquare from '../../InfoSquare'
...
 import identityIconSvg from '../../images/identityIconSvg'
 import CommonFriends from '../../CommonFriends'
-import {atom, useAtomValue} from 'jotai'
+import {atom, useAtomValue, useSetAtom} from 'jotai'
 import createChatStatusAtom from '../../../state/chat/atoms/createChatStatusAtom'
+import Info from '../../Info'
+import {askAreYouSureActionAtom} from '../../AreYouSureDialog'
+import {pipe} from 'fp-ts/function'
+
+const showCommonFriendsExplanationUIAtom = atom(
```
kaladivo: We use actionAtom suffix to indicate only setter is meant to be used in atom.

apps/mobile/src/components/Info.tsx
kaladivo, line 1: It's a great idea to create general element for this. 

How about handling displayed state inside this component? The close button will always make this invisible. Why not just do something like: 

```
const [isVisible, setVisible] = useState(true)

...
if(!isVisible) return null

<TouchableOpacity onPress={() => {setVisible(false)}}>
     <SvgImage stroke={tokens.color.pink.val} source={closeSvg} />
</TouchableOpacity>
```

Issue #338, opened 2023-06-20
Flag offer does not work


leomira:
This is a feature crucial for an effective content moderation.

Currently it shows "Unknown error" after I confirm the report.

leomira:
@kaladivo this seems to be working now (at least from the FE perspective), the last thing missing is just a confirmation popup.
<img width="196" alt="Snímka obrazovky 2023-06-21 o 11 13 47" src="https://github.com/vexl-it/vexl/assets/25779015/e3822d0e-a488-44da-81be-754d0aff0809">

leomira:
fixed 🎉

PR #348, opened 2023-06-22
bugfix: #338 - Offer report confirm dialog

SamTremko:


apps/mobile/src/components/OfferDetailScreen/api.ts
```diff
@@ -79,19 +83,35 @@ export function useReportOfferHandleUI(): (
...
         }),
+        TE.mapLeft((e) => {
+          if (e._tag !== 'UserDeclinedError') {
+            Alert.alert(toCommonErrorMessage(e, t) ?? t('common.unknownError'))
+          }
+          loadingOverlay.hide()
+          return false
+        }),
+        TE.map(() => {
+          store.set(reportedFlagAtom, true)
```
kaladivo, line 94: Let's move `loadingOverlay.hide()` here. After this is run, there is no loading happening on the background. It does make sense to remove the overlay as soon as possible after loading is done.

apps/mobile/src/components/OfferDetailScreen/api.ts
```diff
@@ -79,19 +83,35 @@ export function useReportOfferHandleUI(): (
...
+        }),
+        TE.chainFirstW(() => {
```
kaladivo: It might be a good practice to change this to `TE.chainFirstTaskK`, it does not matter if TE results in Left value, right? With this code, it would not hide loadingOverlay, also it would not go back, since in the following match, it would trigger onLeft.

kaladivo: Left value could be retuned if user presses back button for example.

kaladivo (changes requested):

apps/mobile/src/components/OfferDetailScreen/api.ts
SamTremko: As there will not be any "negative" button present in the dialog it should not happen that on left is triggered, but this is definitely cleaner solution

SamTremko: But I get it that in the future the layout can change, so this way we will not have any problems

kaladivo (approved):

Issue #274, opened 2023-01-31
Allow users to report/flag the offer directly from the chat. 


zaplmi:
It would be grate if user could report/flag the offer directly from the chat after being connected with the counterparty (after realising or experiencing that the offer/user who posted it is shady or scammy). 

The offer in the marketplace gets greyed out and such option is not available. 

Currently, the only option is just to block the user in the chat.

leomira:
Makes total sense and actually this is a really easy from a UX perspective with the current UI, would just add it here, below "Reveal identity" and a flag button next to block button on the top:
![signal-2023-06-13-220644](https://github.com/vexl-it/vexl/assets/25779015/e390defd-298b-4fe9-b9b5-3cb9f3a53baf)

SamTremko:
@leomira I implemented it like this:


![Image](https://github.com/vexl-it/vexl/assets/32439382/a570af9a-b004-4494-a268-f5be896ea3a8)

and if offer was already reported and chat is still available

![Image](https://github.com/vexl-it/vexl/assets/32439382/31fdf5dc-c845-4f65-b591-313b7565c48a)

leomira:
@SamTremko za mna parada!

leomira:
@SamTremko works fine, just one small issue: when offer is reported via chat, the number of total amount of offers does not update on the marketplace

leomira:
now it seems to be updating the total count!

PR #352, opened 2023-06-24
feate: #274 - Report offer from chat

SamTremko:


apps/mobile/src/components/OfferDetailScreen/components/OfferInfo.tsx
```diff
@@ -43,13 +43,15 @@ function OfferInfo({offer}: {offer: OneOfferInState}): JSX.Element {
...
     <Stack f={1} mx={'$2'} my={'$4'}>
       <ScreenTitle text={t('offer.title')}>
-        <IconButton
-          variant={'dark'}
-          icon={flagSvg}
-          onPress={() => {
-            void reportOffer(offer.offerInfo.offerId)()
-          }}
-        />
+        {!offer.flags.reported && (
```
kaladivo, line 46: I understand this is a good practice to do. But when can offer be reported and at the same time displayed? That should not happen right?

kaladivo (approved):

apps/mobile/src/components/OfferDetailScreen/components/OfferInfo.tsx
SamTremko: This can happen when user does like so:
Request offer -> Go to detail -> Go to chat -> report offer -> after reporting it goes back to offer detail -> and there should not be the flag option available (maybe this is sth to discuss if we should not redirect in all cases to marketplace)

PR #356, opened 2023-06-26
fix: 1st level offer should not encrypt to 2nd lvl connections

kaladivo:
How to test: 

1. Create new 1st lvl offer with code from main branch
2. while still in the main branch, put app on background and open it again
3. Offer is now encrypted for 2nd lvl friends although it was intended for 1st lvl only
4. Checkout this branch
5. run `yarn turbo:build` - to build all changes
6. Rerun the app on the same device with the same state from steps 1-3
7. Offer should now be encrypted only for 1st lvl friends 2nd lvl private parts should be removed from the server


kaladivo:
I've tested this and it works. But please, just to make sure, test it also @SamTremko. Steps are outlined above. If it works, please accept this PR

SamTremko (approved):

Issue #361, opened 2023-06-28
Different number of common friends in offer detail


SamTremko:
<img width="400" alt="image" src="https://github.com/vexl-it/vexl/assets/32439382/799ac66b-50ac-438a-9627-8fbc83b051e2">

leomira:
works fine!

PR #366, opened 2023-06-28
fix: #361 - Different number of common friends in offer detail

SamTremko:


apps/mobile/src/components/OfferAuthorAvatar.tsx
```diff
@@ -21,8 +23,16 @@ function OfferAuthorAvatar({
   const chatForOffer = useChatForOffer({
     offerPublicKey: offerInfo.publicPart.offerPublicKey,
   })
+  const store = useStore()
   const userData = useAtomValue(userDataAtom)
   const {t} = useTranslation()
+  const commonFriends = useMemo(
+    () =>
+      store.get(
```
kaladivo, line 31: Is there any reason why use `store.get` instead of `useAtomValue` here?

kaladivo (approved):

apps/mobile/src/components/OfferAuthorAvatar.tsx
SamTremko: You are right, there is no reason for that

Issue #293, opened 2022-09-24
Avatar in Reveal identity is too tiny (and cannot be enlarged)


slush0:
Avatar in Reveal identity (and later in chat window) is too tiny and cannot be enlarged. That may be problem for some photos where the person is taken from distance, with bad lightning etc.

kaladivo:
AC: 
- when user presses on reveal identity message's avatar it opens it in full screen view `
- There is close icon at the top right of the full screen view of the avatar
- When user presses back (on android) full avatar closes

PR #372, opened 2023-07-07
feature: #293 - Avatar can be enlarged and zoomed

SamTremko:


apps/mobile/src/components/ChatDetailScreen/components/MessagesScreen.tsx
```diff
@@ -49,6 +50,7 @@ function MessagesScreen(): JSX.Element {
           <ChatTextInput />
         </Stack>
       )}
+      <UserAvatarDetail />
```
kaladivo: This is a good place to put this 👍

apps/mobile/src/state/chat/atoms/showUserAvatarDetailAtom.ts
```diff
@@ -0,0 +1,4 @@
+import {atom} from 'jotai'
+import {type UriString} from '@vexl-next/domain/dist/utility/UriString.brand'
+
+export const userAvatarDetailAtom = atom<UriString | undefined>(undefined)
```
kaladivo: Is there any reason why is this living here, instead of in molecule in `ChatDetailScreen/atoms/index.ts`? It makes more sense, since this is not a global state rather it is a state related to one chat.

apps/mobile/src/components/ChatDetailScreen/components/UserAvatarDetail.tsx
```diff
@@ -0,0 +1,66 @@
...
+        <Stack pos={'absolute'} top={top} right={'$4'} zIndex={'$100'}>
+          <IconButton
+            icon={closeSvg}
+            onPress={() => {
+              setUserAvatarDetail(undefined)
+            }}
+          />
+        </Stack>
+        {userAvatarDetail && (
+          <ImageZoom
```
kaladivo: Zoom does not work on android emulator (holding _cmd_ to use pinch gesture). Am I doing something wrong?

apps/mobile/src/components/ChatDetailScreen/components/UserAvatarDetail.tsx
```diff
@@ -0,0 +1,66 @@
...
+import {useAtom} from 'jotai'
+import {Stack} from 'tamagui'
+import {useSafeAreaInsets} from 'react-native-safe-area-context'
+import IconButton from '../../IconButton'
+import closeSvg from '../../images/closeSvg'
+import {ImageZoom} from '@likashefqet/react-native-image-zoom'
+import {useFocusEffect} from '@react-navigation/native'
+import React from 'react'
+
+function UserAvatarDetail(): JSX.Element {
```
kaladivo: Can you please rename this file so it is apparent this is overlay/modal. I suggest `ImageZoomOverlay`

apps/mobile/src/state/chat/atoms/showUserAvatarDetailAtom.ts
kaladivo: Also, it would be nice, if we could, in the future use this for displaying any image. Not just user avatar. I would suggest renaming this to `openedImageAtom` (and moving this to molecule as mentioned above)

kaladivo (changes requested):

kaladivo:
We talked about option to save the image. I think it might be a good idea to include saving image as a part of this pr. 

it is one liner to share the file (literally - we already have expo-sharing library setup): 

```typescript
import * as Sharing from 'expo-sharing'
...
await Sharing.shareAsync(fileUri). 
```

Share icon can be added next to the close icon.

apps/mobile/src/state/chat/atoms/showUserAvatarDetailAtom.ts
kaladivo: Resolved. Thanks for moving this to ChatDetailScreen molecule

apps/mobile/src/components/ChatDetailScreen/components/ImageZoomOverlay.tsx
```diff
@@ -0,0 +1,109 @@
...
+import shareAndroidSvg from '../images/shareAndroidSvg'
+import {toError} from '@vexl-next/domain/dist/utility/errors'
+
+const ZoomedImage = gestureHandlerRootHOC(() => {
+  const {width} = useWindowDimensions()
+  const {bottom, top, right, left} = useSafeAreaInsets()
+  const {openedImageUriAtom} = useMolecule(chatMolecule)
+  const [openedImageUri, setOpenedImageUri] = useAtom(openedImageUriAtom)
+
+  async function shareImage(): Promise<void> {
```
kaladivo: It seems wrong to use this without useCallback. But I also don't see any props/hooks that might cause this to rerender (and trigger unnecessary rerendering). But I would still prefere to have this in useCallback. 

Is there any reason why is this not in useCallback?

kaladivo: Also, does this function work?

Why is shareImage function async? How is error being handled?

kaladivo: I would suggest either removing Either.tryCatch and handling this with try catch block, or using TaskEither and mapping left to catch errors.

kaladivo (approved):

PR #373, opened 2023-07-10
Reintroduce local symbols used for currencies

prusnak:
Followup to https://github.com/vexl-it/vexl/pull/336 which restores some of the functionality of local currency symbols

Please test the PR before merging this - I haven't tested it


apps/mobile/src/utils/localization/currency.ts
```diff
@@ -0,0 +1,80 @@
+import {bigNumberToString} from '../bigNumberToString'
+
+interface CurrencyInfo {
+    code: string                  // ISO-4217
```
kaladivo: It would be good practice to make these properties readonly. 
Also, we are using branded types in the codebase, I would much more prefer defining the interface like this (in currency.brand.ts): 

```typescript
import {z} from 'zod'
export const Currency = z.object({
    code: z.string().brand<'currencyCode'>(), // it can be further refined but that, I don't think is necessary in this case.
    flag: z.string().brand<'currencyFlag'>(),
    name: z.string().brand<'currencyName'>(),
    symbol: z.string().brand<'currencySymbol'>(),
    position: z.enum(['before', 'after']),
})
export type Currency = z.TypeOf<typeof Currency>
```

If we do it that way, and later in the code we write a function that accepts flag. We can force the linter to only accept `Currency['flag']` type instead of generic `string` - since the only place where those brands are created is in this file (when defining `currencies` constant), it is very hard to pass value that we did not intend to. This saves a lot of headache later on ;).

apps/mobile/src/utils/localization/currency.ts
```diff
@@ -0,0 +1,80 @@
...
+    flag: string                  // flag emoji
+    name: string                  // currency name
+    symbol: string                // local symbol
+    position: 'before' | 'after'  // local symbol position
+}
+
+export const currencies: Record<string, CurrencyInfo> = {
```
kaladivo: Do we need to export this?

kaladivo: Also, If we use the currency definition as in previous comment. We can rewrite the type definition like so: 

`const currencies: Record<Currency['code'], Currency>`. This way linter will allow us to use only keys that exists.

For fun purposes, let's imagine we don't use Zod (or branded types) in the project. It would be much better to omit explicit type definition and just put `as const` at the end of assignment, this would also make typescript aware of possible key values so it won't allow accessing any `string` key but only those defined.

apps/mobile/src/utils/localization/currency.ts
```diff
@@ -0,0 +1,80 @@
...
+    SGD: {code: 'SGD', flag: '🇸🇬', name: 'Singapore dollar', symbol: 'S$', position: 'before'},
+    THB: {code: 'THB', flag: '🇹🇭', name: 'Thai baht', symbol: '฿', position: 'before'},
+    TRY: {code: 'TRY', flag: '🇹🇷', name: 'Türk lirası', symbol: '₺', position: 'before'},
+    UAH: {code: 'UAH', flag: '🇺🇦', name: 'Українська гривня', symbol: '₴', position: 'after'},
+    USD: {code: 'USD', flag: '🇺🇸', name: 'US dollar', symbol: '$', position: 'before'},
+    ZAR: {code: 'ZAR', flag: '🇿🇦', name: 'South African rand', symbol: 'R', position: 'before'},
+}
+
+export function getCurrencyLeftText(currencyCode: string | undefined): string | undefined {
+    if (currencyCode === undefined) {
```
kaladivo: Pro tip ⭐️. We could rewrite this to be shorter and easier to read with optional chaining operator: 

```typescript
export function getCurrencyLeftText(currencyCode: string | undefined): string | undefined {
    const currency = currencies[currencyCode]
    if (currency?.position !== 'before') return undefined

    return currency.symbol;
 }
```

kaladivo (changes requested):
Let me know if you want to implement the changes or if me or Sam should take over and finish this.

apps/mobile/src/utils/localization/currency.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+    if (currencies[currencyCode].position !== 'after') {
+        return undefined;
+    }
+    return currencies[currencyCode].symbol;
+}
+
+export function formatCurrencyAmount(currencyCode : string, amount: number): string {
+    const currency = currencies[currencyCode];
+    if (currency.position === 'before') {
```
kaladivo: We also need to use optional chaining operator here (`currency?.position` === `before`) operator here otherwise this will throw an error trying to access property of undefined (ofc with branded types, we don't have to care about this ;) - that is the case only if this function does not accept nullable value of `currencyCode`).

apps/mobile/src/utils/localization/currency.ts
kaladivo: And with branded types, it would look something like this: 


```typescript
export function getCurrencyLeftText(currencyCode: Currency | undefined): string | undefined {
     if (!currencyCode) return undefined;

    const currency = currencies[currencyCode] 

    if (currency.position !== 'before') return undefined // here, it will always have value.

    return currency.symbol;
 }
```

apps/mobile/src/utils/localization/currency.ts
```diff
@@ -0,0 +1,80 @@
...
+    if (currencies[currencyCode] === undefined) {
+        return undefined;
+    }
+    if (currencies[currencyCode].position !== 'before') {
+        return undefined;
+    }
+    return currencies[currencyCode].symbol;
+}
+
+export function getCurrencyRightText(currencyCode: string | undefined): string | undefined{
```
kaladivo: 👍 I really like that the parameter type is not being defined as this: 
```getCurrencyRightText(currencyCode?: string): string | undefined```

But rather with `| undefined`. 

It makes it apparent that the function expects to receive some value in parameter rather than being called without it.

**Comment by prusnak** [comment]
I think that all outstanding points have been addressed https://github.com/vexl-it/vexl/pull/373/commits/ab128464f90dbf32e89e6a3b2697e519dec10651

Added one more commit https://github.com/vexl-it/vexl/pull/373/commits/151f6793b0cbf267566f166fc31b67667691d6fb that renames `Currency` to `CurrencyCode` where applicable, so now only the components in apps/mobile/src/components/... are named Currency, not the underlying data classes.

File: apps/mobile/src/components/OfferForm/components/Description.tsx [hunk 6]
```diff
@@ -23,6 +23,7 @@ function Description({offerDescriptionAtom}: Props): JSX.Element {
       </Text>
       <Stack mt="$4" br="$4" p="$4" bc="$grey">
         <Input
+          autoCorrect={false}
```
kaladivo: Leave this enabled for users

kaladivo (approved):

Issue #319, opened 2023-06-15
Add more translations


leomira:
With the international launch in mind, we should support more translations.
German, Italian, French and Spanish seem to be the natural choice. 
Automatic translations followed by a review by a native speaker should do the trick.

Connected to #267

leomira:
+ portuguese

kaladivo:
connected to #257

PR #375, opened 2023-07-12
#319 - add more translations, add option to change language in dev mode

kaladivo:


packages/localizations/src/chunks/other.cs.ts
```diff
@@ -236,7 +236,7 @@ const otherCs: typeof en =
       "filterOffers": "Filtrovat nabídky",
       "numberOfCommon": "{{number}} společné",
       "offerNotFound": "Nabídka nebyla nalezena. Možná byla smazána autorem",
-      "inputPlaceholder": "Tady napiš zprávu...",
```
SamTremko, line 239: In this case we would need this translation: `"inputPlaceholder": "Tady napiš zprávu..."`,
I've changed it few sprints ago

packages/localizations/src/chunks/other.sk.ts
```diff
@@ -235,7 +235,7 @@ const otherSk: typeof en =
       "filterOffers": "Filtrovanie ponúk",
       "numberOfCommon": "{{number}} spoločné",
       "offerNotFound": "Ponuka nebola nájdená. Autor ju možno vymazal",
-      "inputPlaceholder": "Tu napíš správu...",
```
SamTremko: The same here, we need: `"inputPlaceholder": "Tu napíš správu..."`

packages/localizations/src/chunks/other.en.ts
```diff
@@ -234,7 +234,7 @@ const otherEn =
       "numberOfCommon": "{{number}} common",
       "offerNotFound":
         "Offer not found. It might have been deleted by the author",
-      "inputPlaceholder": "Let’s write a message here...",
```
SamTremko, line 237: And the same here, we need: `"inputPlaceholder": "Let’s write a message here..."`

packages/localizations/utilsSrc/newTranslation.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export default async function newTranslation(): Promise<void> {
+  const tableContent = await getTranslations()
+
+  const keys = readFileSync('out/keys', 'utf-8').split('\n')
+  const en = readFileSync('out/en', 'utf-8').split('\n')
+
+  const de = readFileAndReplaceTemplate('out/de', en)
+
+  console.log(de)
```
SamTremko: Do we need this `console.log` here?

SamTremko (approved):

packages/localizations/src/chunks/other.cs.ts
kaladivo: If you want to update existing translations, you always need to do that in the table otherwise it will be overridden in the next sync. I have updated the table

packages/localizations/utilsSrc/newTranslation.ts
kaladivo: We don't, my bad. Good catch!

Issue #225, opened 2023-06-07
Pressing the same currency as previously selected resets the slider in create offer


SamTremko:
Steps to reproduce:

1. Go to Create Offer
2. Move the slider
3. Press the same currency sign again
4. Slider is reset

PR #385, opened 2023-07-18
bugfix: #225 - Pressing the same currency does not reset limits

SamTremko:


apps/mobile/src/components/ModifyOffer/atoms/offerFormStateAtoms.ts
```diff
@@ -112,14 +112,17 @@ export const offerFormMolecule = molecule(() => {
...
-      set(currencyAtom, currency)
-      set(amountBottomLimitAtom, get(amountBottomLimitUsdEurCzkAtom))
-      set(
-        amountTopLimitAtom,
-        currency === 'CZK'
-          ? get(amountTopLimitCzkAtom)
-          : get(amountTopLimitUsdEurAtom)
-      )
+      const currencyFromAtom = get(currencyAtom)
+      if (currencyFromAtom !== currency) {
```
kaladivo: It's nice practice to actually invert conditions like this to make the code more readable and prevent unnecessary nesting and the pyramid of doom [see this](https://www.youtube.com/watch?v=CFRhGnuXG-4)

in this case that would be: 
`if(previousCurrency === newCurrency) return`

kaladivo (approved):
Please look at comments before merging.

apps/mobile/src/components/ModifyOffer/atoms/offerFormStateAtoms.ts
SamTremko: @kaladivo good point! thx

PR #411, opened 2023-08-10
fix: improvements

AdamSchinzel:
Bunch of minor improvements:

1. Clear button in input is visible only if the input has value 
1. Bottom buttons on App logs screen are disabled if there are no app logs
1. Fixed dividers around currency section in settings (if `enableHiddenFeatures` was `false` it looked weird, see screenshot)
1. Cancel button in alert when selecting profile photo has on iOS cancel button styling

IMG_1485


apps/mobile/src/components/AppLogsScreen/index.tsx
```diff
@@ -12,19 +12,21 @@ import {
...
 import closeSvg from '../images/closeSvg'
 import useSafeGoBack from '../../utils/useSafeGoBack'
+import { appLogAtomsAtom } from './atoms'
 
 function AppLogsScreen(): JSX.Element {
   const {t} = useTranslation()
   const [enabled, setEnabled] = useState(getCustomLoggingEnabled())
   const setLoading = useSetAtom(loadingOverlayDisplayedAtom)
   const safeGoBack = useSafeGoBack()
+  const logsAtoms = useAtomValue(appLogAtomsAtom)
```
kaladivo: This will cause unnecessary rerender of this component when new logs are written. It would be much better to use a selectAtom like: 

```
const appLogsEmptyAtom = selectAtom(appLogsAtom, logs => logs.length === 0)
```

apps/mobile/src/components/InsideRouter/components/SettingsScreen/components/ButtonsSection.tsx
```diff
@@ -158,6 +158,42 @@ function ButtonsSection(): JSX.Element {
             children: <ContactsImportedTitle />,
           },
         ],
+        enableHiddenFeatures
```
kaladivo, line 162: Good idea to add this.

kaladivo (changes requested):

apps/mobile/src/components/AppLogsScreen/index.tsx
AdamSchinzel: 9f3bf36bdb36ae3eb8d540313107c9da459dd4bc

apps/mobile/src/components/AppLogsScreen/index.tsx
```diff
@@ -12,20 +12,24 @@ import {
...
 import showErrorAlert from '../../utils/showErrorAlert'
+import { appLogAtomsAtom } from './atoms'
+import { selectAtom } from 'jotai/utils'
 
 function AppLogsScreen(): JSX.Element {
   const {t} = useTranslation()
   const [enabled, setEnabled] = useState(getCustomLoggingEnabled())
   const setLoading = useSetAtom(loadingOverlayDisplayedAtom)
   const safeGoBack = useSafeGoBack()
+  const appLogsEmptyAtom = selectAtom(appLogAtomsAtom, logs => logs.length === 0)
```
kaladivo: If possible, always create atoms outside of component render method. Creating the atom here will make the selector function run every time component is rendered and possibly trigger unnecessary rerenders.

You can add the atom to ./atoms file.

kaladivo (changes requested):

apps/mobile/src/components/AppLogsScreen/index.tsx
AdamSchinzel: 86f28794c4d45782f489bd9e4e30c64d3ba7a8c0

kaladivo (approved):

PR #414, opened 2023-08-15
chore: sync translations & up version

kaladivo:


SamTremko:
@kaladivo pls re-sync translations once more, I did some corrections in german texts during the last hours, thx!

PR #437, opened 2023-08-28
chore: Add apk building workflow

kaladivo:


.github/workflows/build-eas-staging-apk.yaml
```diff
@@ -0,0 +1,27 @@
+name: "[Build] Build staging APK"
```
vdovhanych: second build is redundant "[Build] staging APK" is sufficient

vdovhanych (approved):

SamTremko (approved):

Issue #447, opened 2023-09-04
Location search screen does not clear previous results when search term is changed


kaladivo:
See video for reproduce instruction. 

In the last query `brno` returns only one item from server, but the previous search results are still displayed.

https://github.com/vexl-it/vexl/assets/11568934/ad6ceb70-4752-4741-a2c1-078ddb99db20

PR #467, opened 2023-09-06
fix: #447 - Location screen does not clear previous results

SamTremko:


SamTremko:
Closes #447

apps/mobile/src/components/FilterOffersScreen/atom.ts
```diff
@@ -114,6 +121,46 @@ export const amountTopLimitAtom = focusAtom(offersFilterAtom, (optic) =>
   optic.prop('amountTopLimit')
 )
 
+export const locationSuggestionsAtom = atom<LocationSuggestion[]>([])
+
+export const locationSuggestionsAtomsAtom = splitAtom(locationSuggestionsAtom)
+
+export const getLocationSuggestionsActionAtom = atom(
```
kaladivo: The name is missleading, this refreshes locationSuggestionsAtom. How about renaming it to something like `updateAndRefreshLocationSuggestionsActionAtom` or `triggerLocationSuggestionsQueryActionAtom`)

apps/mobile/src/components/FilterOffersScreen/components/LocationSearch/LocationCell.tsx
```diff
@@ -1,24 +1,39 @@
...
+import {TouchableOpacity} from 'react-native'
+import {type LocationSuggestion} from '@vexl-next/rest-api/dist/services/location/contracts'
+import {type Atom, useAtomValue, useSetAtom} from 'jotai'
+import {setOfferLocationActionAtom, locationSuggestionsAtom} from '../../atom'
 
-interface Props extends TouchableOpacityProps {
-  city: string
-  country: string
+interface Props {
+  locationSuggestionAtom: Atom<LocationSuggestion>
```
kaladivo, line 10: Nice! Good idea to rename this👍

apps/mobile/src/components/FilterOffersScreen/components/LocationSearch/LocationsList.tsx
```diff
@@ -0,0 +1,44 @@
...
+  onClose,
+}: Props): JSX.Element {
+  return (
+    <Stack pt={'$2'}>
+      <FlatList
+        showsHorizontalScrollIndicator={false}
+        showsVerticalScrollIndicator={false}
+        keyExtractor={atomKeyExtractor}
+        data={locationSuggestionsAtoms}
+        renderItem={({item: locationSuggestionAtom}) =>
```
kaladivo: This neglects the purpouse of defining the `renderItem` function outside of the component. We need the props to stay equal between renders, but here, `renderItem` receives different function on every render. 

In this case, it might be a good idea to use `useMemo` . 

See [this](https://stackoverflow.com/questions/74333190/react-native-whole-flatlist-rerenders-on-every-change-even-with-react-memo) for further explanation on why this is a good idea to do.

apps/mobile/src/components/FilterOffersScreen/components/LocationSearch/index.tsx
```diff
@@ -1,98 +1,60 @@
...
+    if (value === '') setLocationSuggestions([])
+    setInputValue(value)
   }
 
   useEffect(() => {
     if (debouncedSearchValue.length > 0) {
-      searchLocation()
+      void getLocationSuggestions({
+        phrase: debouncedSearchValue,
+        lang: i18n.locale,
```
kaladivo, line 45: This is great 👍

apps/mobile/src/components/FilterOffersScreen/components/LocationSearch/index.tsx
```diff
@@ -1,98 +1,60 @@
...
-        (locations) => {
-          const locationWithMunicipality = locations.result.filter(
-            (location) => location.userData.municipality !== ''
-          )
-          setLocationResults({result: locationWithMunicipality})
-        }
-      )
-    )()
-  }, [t, debouncedSearchValue, getLocationSuggestions])
+  const getLocationSuggestions = useSetAtom(getLocationSuggestionsActionAtom)
```
kaladivo: This is also a great idea to extract the logic into atoms 👍

apps/mobile/src/components/FilterOffersScreen/components/LocationSearch/index.tsx
```diff
@@ -108,31 +70,23 @@ function LocationSearch({
         <Input
           autoFocus
           value={inputValue}
-          onChangeText={setInputValue}
+          onChangeText={(value) => {
+            onInputValueChange(value)
```
kaladivo: Better to not create anonymous function when it's not needed: `onChangeText={onInputValueChange}`

packages/rest-api/src/services/location/contracts.ts
```diff
@@ -19,11 +19,14 @@ export const LocationData = z.object({
 })
 export type LocationData = z.TypeOf<typeof LocationData>
 
-export const LocationSuggestions = z.object({
+export const LocationSuggestion = z.object({
```
kaladivo, line 22: Good catch! 😎

apps/mobile/src/state/location/atoms/fetchLocationSuggestionsAtom.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export const fetchLocationSuggestionsAtom = atom(
+  null,
+  (get, set, request: GetLocationSuggestionsRequest) => {
+    const api = get(publicApiAtom)
+    const {t} = get(translationAtom)
+
+    return pipe(
+      api.location.getLocationSuggestions(request),
+      TE.match(
```
kaladivo: You can use `TE.getOrElse` when you don't need to map the right value.

apps/mobile/src/components/OfferForm/components/LocationSearch/LocationsList.tsx
```diff
@@ -0,0 +1,67 @@
...
+}: Props): JSX.Element {
+  return (
+    <Stack pt={'$2'}>
+      <FlatList
+        showsHorizontalScrollIndicator={false}
+        showsVerticalScrollIndicator={false}
+        keyExtractor={atomKeyExtractor}
+        data={locationSuggestionsAtoms}
+        renderItem={({item: locationSuggestionAtom}) =>
+          renderItem(
```
kaladivo: I wrote about this above.

kaladivo (changes requested):
Do we really need to have LocationSearch modal defined twice? Is the usecase of these two components differs so much it would be too complicated to merge them into one?

kaladivo (approved):

kaladivo (approved):

PR #469, opened 2023-09-06
feat: small-fixes

SamTremko:


apps/mobile/src/components/FilterOffersScreen/components/LocationSearch/index.tsx
```diff
@@ -94,19 +95,38 @@ function LocationSearch({
     }
   }, [debouncedSearchValue, searchLocation])
 
+  useEffect(() => {
+    const timeout = setTimeout(() => {
```
kaladivo: Should this be here? What if component is mounted but the `visible` is false? Wouldn't this just focus the input even if it is not visible?

kaladivo (changes requested):

apps/mobile/src/components/FilterOffersScreen/components/LocationSearch/index.tsx
SamTremko: sry, my bad, forgot to delete it after I moved it to Modals onShow() func

kaladivo (approved):

PR #480, opened 2023-09-11
Feat/#467 remove message of type offer deleted

kaladivo:


apps/mobile/src/components/ChatDetailScreen/components/QuickActionBanner.tsx
```diff
@@ -129,35 +129,36 @@ function QuickActionBanner(): JSX.Element | null {
...
-        onButtonPress={() => {
-          Keyboard.dismiss()
-          void deleteChat().then((result) => {
-            if (result) resetNavigationToMessagingScreen()
-          })
-        }}
-      />
-    )
-  }
+  // if (
```
vdovhanych: Why leave it commented in here and not just delete it?

vdovhanych (approved):

apps/mobile/src/components/ChatDetailScreen/components/QuickActionBanner.tsx
kaladivo: Good point, removing

kaladivo: Glad you asked. This one should have stayed uncommented🥔

PR #490, opened 2023-09-16
fix: make fulltext search less fuzzy

kaladivo:


apps/mobile/src/state/marketplace/utils/filterOffersByText.tsx
```diff
@@ -2,6 +2,7 @@ import {type OneOfferInState} from '@vexl-next/domain/dist/general/offers'
 import {type ContactNormalizedWithHash} from '../../contacts/domain'
 import {matchSorter} from 'match-sorter'
```
prusnak: we can drop this matchSorter import from here

prusnak (changes requested):

apps/mobile/src/state/marketplace/utils/filterOffersByText.tsx
kaladivo: Done

kaladivo: Removed. Sorry for missing this.

prusnak (approved):

PR #499, opened 2023-09-29
Fix/contact multiplicating

kaladivo:


apps/mobile/src/components/ContactListSelect/index.tsx
```diff
@@ -35,6 +39,14 @@ function ContactsListSelect({
   const toDisplay = useAtomValue(contactsToDisplayAtomsAtom)
   const loading = useAtomValue(contactsLoadingAtom)
   const submit = useSetAtom(submitActionAtom)
+  const triggerContactsReload = useSetAtom(triggerContactsReloadAtom)
+
+  useOnFocusAndAppState(
+    useCallback(() => {
+      console.log('triggering')
```
SamTremko: pls remove log if it is not needed

SamTremko (approved):

apps/mobile/src/components/ContactListSelect/index.tsx
kaladivo: Removed

PR #500, opened 2023-09-29
fix: Offers refresh triggered periodically

SamTremko:


apps/mobile/src/components/InsideRouter/components/MarketplaceScreen/components/EmptyListPlaceholder.tsx
```diff
@@ -104,7 +106,11 @@ function EmptyListPlaceholder(): JSX.Element {
   }
 
   useEffect(() => {
-    if (reachNumber >= REACH_NUMBER_THRESHOLD && lastImportOfContacts) {
+    if (
+      reachNumber >= REACH_NUMBER_THRESHOLD &&
+      lastImportOfContacts &&
+      offersToSeeInMarketplace.length === 0
```
kaladivo: How about creating select atom: `const isMarketplaceEmptyAtom = selectAtom(offersToSeeInMarketplace, offers => offers.length === 0)`

kaladivo: This will disable refreshing marketplace when just one section (buying or selling) is empty (and the other one has offers). 

How about using `useFocusEffect` instead of useEffect. That should start the interval only when EmptyListPlaceholder is focused.

kaladivo (changes requested):

apps/mobile/src/components/InsideRouter/components/MarketplaceScreen/components/EmptyListPlaceholder.tsx
SamTremko: @kaladivo yeah you are right, I have overthought this one a little ... useFocusEffect solves the issue completely without need of adding another condition

kaladivo (approved):

PR #502, opened 2023-09-29
fix: Price fluctuates between two values

SamTremko:


apps/mobile/src/components/InsideRouter/atoms.ts
```diff
@@ -1,13 +1,20 @@
...
 import {type Task} from 'fp-ts/Task'
 import reportError from '../../utils/reportError'
 import {type GetCryptocurrencyDetailsResponse} from '@vexl-next/rest-api/dist/services/user/contracts'
 import showErrorAlert from '../../utils/showErrorAlert'
+import {IsoDatetimeString} from '@vexl-next/domain/dist/utility/IsoDatetimeString.brand'
+import {DateTime} from 'luxon'
+
+const BTC_PRICE_UPDATE_TRIGGER_THRESHOLD_MINUTES = 15
+
+const btcPriceLastUpdateAtAtom = atom<IsoDatetimeString | undefined>(undefined)
```
kaladivo: UnixMilliseconds can be used instead to make time operations more clear.

apps/mobile/src/components/InsideRouter/atoms.ts
```diff
@@ -18,6 +25,20 @@ export const refreshBtcPriceActionAtom = atom<undefined, [], Task<boolean>>(
   (get, set) => {
     const api = get(privateApiAtom)
     const {t} = get(translationAtom)
+    const btcPriceLastUpdateAt = get(btcPriceLastUpdateAtAtom)
+    const timeDifferenceSinceLastUpdate = Math.round(
```
kaladivo: Can you wrap this in `if(btcpriceLastUpdateAt)`? (or use `UnixMilliseconds` and make it non nullable with default 0 so it always runs with initial value).

kaladivo (changes requested):

apps/mobile/src/components/InsideRouter/atoms.ts
```diff
@@ -18,6 +27,19 @@ export const refreshBtcPriceActionAtom = atom<undefined, [], Task<boolean>>(
...
+      DateTime.now().diff(
+        DateTime.fromMillis(btcPriceLastUpdateAt ?? Date.now()),
```
kaladivo: No need to convert to DateTime anymore. You can subtract milliseconds directly ;).

Also, remove `?? Date.now()` otherwise it won't trigger the update with initial value (the following condition will be truthy).

kaladivo (changes requested):

apps/mobile/src/components/InsideRouter/atoms.ts
```diff
@@ -18,6 +27,16 @@ export const refreshBtcPriceActionAtom = atom<undefined, [], Task<boolean>>(
   (get, set) => {
     const api = get(privateApiAtom)
     const {t} = get(translationAtom)
+    const btcPriceLastUpdateAt = get(btcPriceLastUpdateAtAtom)
+    const timeDifferenceSinceLastUpdate = Math.round(
```
kaladivo: This is also a good solution. I would prefer to use miliseconds and `BTC_PRICE_UPDATE_TRIGGER_THRESHOLD_MILISECONDS` instead. 

Is `Math.round` really needed in this case? Let's not complicate the code unnecessarily.

apps/mobile/src/components/InsideRouter/atoms.ts
```diff
@@ -32,6 +51,7 @@ export const refreshBtcPriceActionAtom = atom<undefined, [], Task<boolean>>(
           return false
         },
         (r) => {
+          set(btcPriceLastUpdateAtAtom, UnixMilliseconds.parse(Date.now()))
```
kaladivo: FYI there is already method for this: `unixMillisecondsNow` in `@vexl-next/domain/dist/utility/UnixMilliseconds.brand`

apps/mobile/src/components/InsideRouter/atoms.ts
```diff
@@ -1,13 +1,22 @@
...
 import reportError from '../../utils/reportError'
 import {type GetCryptocurrencyDetailsResponse} from '@vexl-next/rest-api/dist/services/user/contracts'
 import showErrorAlert from '../../utils/showErrorAlert'
+import {Duration} from 'luxon'
+import {UnixMilliseconds} from '@vexl-next/domain/dist/utility/UnixMilliseconds.brand'
+
+const BTC_PRICE_UPDATE_TRIGGER_THRESHOLD_MINUTES = 15
+
+const btcPriceLastUpdateAtAtom = atom<UnixMilliseconds>(
+  UnixMilliseconds.parse(0)
```
kaladivo: There is constant for this `UnixMilliseconds0` in `@vexl-next/domain/dist/utility/UnixMilliseconds.brand`.

kaladivo (approved):

PR #514, opened 2023-10-03
Honeybadger does not care

prusnak:
This is a draft PR that removes all Bitcoin references from the code (only from the code, the localization strings / other texts such as ToS still use BTC/Bitcoin) and replaces them with the word honey. That means we might want to decide to merge this even when we'd like to keep the word Bitcoin in the user interface.

This PR is just to see what kind of changes would be needed.

### DIFF
--- a/apps/mobile/src/components/InsideRouter/atoms.ts
+++ b/apps/mobile/src/components/InsideRouter/atoms.ts
@@ -7,7 +7,7 @@ import {toCommonErrorMessage} from '../../utils/useCommonErrorMessages'
 import {translationAtom} from '../../utils/localization/I18nProvider'
 import {type Task} from 'fp-ts/Task'
 import reportError from '../../utils/reportError'
-import {type GetCryptocurrencyDetailsResponse} from '@vexl-next/rest-api/dist/services/user/contracts'
+import {type GetHoneyDetailsResponse} from '@vexl-next/rest-api/dist/services/user/contracts'
 import showErrorAlert from '../../utils/showErrorAlert'
 import {
   type UnixMilliseconds,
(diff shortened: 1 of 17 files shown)


prusnak:
There is one open issue that needs to be addressed:

```
$ yarn turbo:lint
@vexl-next/mobile-app:lint: src/components/InsideRouter/atoms.ts(41,32): error TS2345: Argument of type '{}' is not assignable to parameter of type '{ coin: string; }'.
@vexl-next/mobile-app:lint:   Property 'coin' is missing in type '{}' but required in type '{ coin: string; }'.
```

And indeed, I removed the coin property from there, however, below I declare the coin property as optional with a default (so the app does not have to fill it in and the server will put the default if not set):

https://github.com/vexl-it/vexl/blob/8fb5d9701885a319a81adaf2ac46b6c6b7ebfb98/packages/rest-api/src/services/user/contracts.ts#L98-L100

Please advise how to fix this issue @kaladivo

kaladivo:
@prusnak 

The problem here, is that validated type != type that can be validated (input of validation != output of validation). There are functions to infer input/output types (`z.input` and `z.output`). In our particular case that means that coin is always defined on validated type but is optional on input type.

App can send this: `interface GetHoneyDetailRequest {coin: string | undefined}` but server wants to have this: `interface GetHoneyDetailRequest {coin: string}`

Now, I would love to have one interface for both, that will reduce complexity so developer does not have to think which type to use for one particular scheme. 

That means defining the type like this: 

```typescript
export const GetHoneyDetailsRequest = z.object({
  coin: z.string()
})
export type GetHoneyDetailsRequest = z.TypeOf<typeof GetHoneyDetailsRequest>
```

And changing the function signature in `packages/rest-api/src/services/user/index.ts`, to be: 
```typescript
getHoneyDetails: (request: GetHoneyDetailsRequest = {coin: 'bitcoin'}) => {...
```

Then the function can be called without any parameter. like so: `api.user.getHoneyDetails(),`

It seems clearer to me. Schema definition does not have to know the default value. 

Also, since server would probably not be able to handle every string in code property, I would suggest using literal or enum, but that might not be viable given the nature of the change we want to make.

See [this](https://github.com/colinhacks/zod/issues/2491) for more info on optional().default() values.

kaladivo:
Doing this also will make any debugging much easier since the request itself will include `coin: 'bitcoin'` parameter, and thus be more transparent.

prusnak:
Let's close this for now.

PR #518, opened 2023-10-04
fix: Wrong colors of report offer image

SamTremko:


kaladivo (changes requested):
Please upload @2 and @3 image sizes so images do not look blurry on devices with higher screen densities. 

See [official documentation](https://reactnative.dev/docs/images)

kaladivo (approved):

PR #558, opened 2023-11-01
fix: Bad chat messages order

SamTremko:


kaladivo (changes requested):
Do not remove `inverted` prop from MessagesList

kaladivo (approved):

PR #566, opened 2023-11-02
fix: Check fonst and text in onboarding for contacts import

SamTremko:


apps/mobile/src/components/LoginFlow/components/AnonymizationAnimationScreen/index.tsx
```diff
@@ -59,7 +59,7 @@ function AnonymizationAnimationScreen({
 
   return (
     <>
-      <HeaderProxy showBackButton={true} progressNumber={undefined} />
+      <HeaderProxy showBackButton={true} progressNumber={1} />
```
kaladivo, line 62: Super, good catch!

kaladivo (approved):

PR #593, opened 2023-11-20
Trade calculator

SamTremko:


apps/mobile/src/components/Info.tsx
```diff
@@ -7,22 +7,35 @@ import Button from './Button'
...
+}
+
+interface PropsWithoutActionButton extends CommonProps {
+  hideActionButton: true
+  actionButtonText?: string
+  onActionPress?: () => void
+}
+
+type Props = PropsWithActionButton | PropsWithoutActionButton
+
```
kaladivo: This introduces unnecessary complexity. Is there reason for defining types like this?

apps/mobile/src/components/PageWithNavigationHeader/state/footerButtonStateAtom.ts
```diff
@@ -0,0 +1,31 @@
...
+  text: undefined,
+  disabled: false,
+  hidden: false,
+  onPress: () => {},
+})
+
+export default footerButtonStateAtom
+
+export function useSetFooterButtonState(
+  set: () => ExtractAtomValue<typeof footerButtonStateAtom>
```
kaladivo: Why not put just a value here? That way we can avoid unnecessary memoization.

apps/mobile/src/components/PageWithNavigationHeader/state/headerStateAtom.ts
```diff
@@ -0,0 +1,31 @@
...
+  hidden: true,
+  title: undefined,
+  goBack: () => undefined,
+  onClose: () => undefined,
+})
+
+export default headerStateAtom
+
+export function useSetHeaderState(
+  getState: () => ExtractAtomValue<typeof headerStateAtom>
```
kaladivo: Why not put just a value here? That way we can avoid unnecessary memoization.

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/atoms.ts
```diff
@@ -0,0 +1,174 @@
...
+  mainTradeCheckListStateAtom,
+  offerForTradeChecklistAtom,
+} from '../../atoms'
+import {publicApiAtom} from '../../../../api'
+import {pipe} from 'fp-ts/function'
+import {AcceptedCurrency} from '@vexl-next/rest-api/dist/services/btcPrice'
+import reportError from '../../../../utils/reportError'
+import {checkInputValueIsDecimal} from '../../utils'
+
+export const SATOSHIS_IN_BTC = 100000000
```
kaladivo: You can always use `_` in number definition to make it more readable. `... = 100_000_000`

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/atoms.ts
```diff
@@ -0,0 +1,174 @@
...
+  null,
+  (get, set, feeAmount: number) => {
+    const btcInputValue = Number(get(btcInputValueAtom))
+    const appliedFee = get(feeAmountAtom)
+
+    const btcValueWithoutFee = btcInputValue / (1 - appliedFee / 100)
+
+    set(
+      btcInputValueAtom,
+      checkInputValueIsDecimal(
```
kaladivo: Rename to make more clear, what this function does.

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/BtcAmountInput.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const ref = useRef<TextInput>(null)
+  const [isFocused, setIsFocused] = useState<boolean>(false)
+
+  const [btcValue, setBtcValue] = useAtom(btcValueAtom)
+  const setFiatValue = useSetAtom(fiatValueAtom)
+  const feeAmount = useAtomValue(feeAmountAtom)
+  const [btcOrSat, setBtcOrSat] = useAtom(btcOrSatAtom)
+  const tradeBtcPrice = useAtomValue(tradeBtcPriceAtom)
+
+  const btcOrSatDropdownData: Array<DropdownItemProps<BtcOrSat>> = useMemo(
```
kaladivo: Does this need to be memoized? Can this be defined outside of the component?

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/BtcAmountInput.tsx
```diff
@@ -0,0 +1,145 @@
...
+        setIsFocused(false)
+      }}
+      onFocus={() => {
+        setIsFocused(true)
+      }}
+      onWrapperPress={() => {
+        ref.current?.focus()
+      }}
+      placeholder={BTC_INPUT_PLACEHOLDER}
+      value={btcValue}
```
kaladivo, line 86: This will result in undefined being passed to input element. It is not recommended to change uncontrolled component to controlled one

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/CalculateAmountScreen/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+    tradePriceTypeDialogVisibleAtom
+  )
+  const syncCalculatedAmountDataStateWithMainState = useSetAtom(
+    syncCalculatedAmountDataStateWithMainStateActionAtom
+  )
+  const saveLocalCalculatedAmountDataStateToMainState = useSetAtom(
+    saveLocalCalculatedAmountDataStateToMainStateActionAtom
+  )
+
+  useEffect(() => {
```
kaladivo, line 46: why not `useFocus` as in proxy elements ?

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/FiatAmountInput.tsx
```diff
@@ -0,0 +1,105 @@
...
+      onFocus={() => {
+        setIsFocused(true)
+      }}
+      onWrapperPress={() => {
+        ref.current?.focus()
+      }}
+      placeholder={`${offerForTradeChecklist?.offerInfo.publicPart.amountTopLimit}`}
+      value={fiatValue}
+      onChangeText={(input) => {
+        calculateBtcValueOnFiatAmountChange(checkInputValueIsDecimal(input))
```
kaladivo: Is it ok to do this on every key press?

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/FiatAmountInput.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const [isFocused, setIsFocused] = useState<boolean>(false)
+
+  const [fiatValue, setFiatValue] = useAtom(fiatValueAtom)
+  const setBtcValue = useSetAtom(btcValueAtom)
+  const feeAmount = useAtomValue(feeAmountAtom)
+  const offerForTradeChecklist = useAtomValue(offerForTradeChecklistAtom)
+  const tradeBtcPrice = useAtomValue(tradeBtcPriceAtom)
+  const btcOrSat = useAtomValue(btcOrSatAtom)
+
+  const calculateBtcValueOnFiatAmountChange = useCallback(
```
kaladivo: Why not make this into atom?

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/PremiumOrDiscountScreen/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const navigation: NavigationProp<TradeChecklistStackParamsList> =
+    useNavigation()
+
+  const offerType = useAtomValue(offerTypeAtom)
+  const feeAmount = useAtomValue(feeAmountAtom)
+  const applyFeeOnFeeChange = useSetAtom(applyFeeOnFeeChangeActionAtom)
+  const [tempFeeAmount, setTempFeeAmount] = useAtom(tempFeeAmountAtom)
+
+  useEffect(() => {
+    setTempFeeAmount(feeAmount)
```
kaladivo: Is fee amount always derived from fee amount? Can `feeAmountAtom` be derived atom?

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/SetYourOwnPriceScreen/components/PriceInfo.tsx
```diff
@@ -0,0 +1,41 @@
...
+interface Props {
+  fiatTempValueAtom: PrimitiveAtom<string | undefined>
+}
+
+function PriceInfo({fiatTempValueAtom}: Props): JSX.Element | null {
+  const {t} = useTranslation()
+  const tradeBtcPrice = useAtomValue(tradeBtcPriceAtom)
+  const fiatTempValue = Number(useAtomValue(fiatTempValueAtom)) ?? 0
+
+  const percentageDifference = useMemo(
```
kaladivo, line 20: Why is this string but later it is always cast to number?

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/SetYourOwnPriceScreen/components/PriceInfo.tsx
```diff
@@ -0,0 +1,41 @@
...
+      hideActionButton
+      text={t(
+        Number(percentageDifference) < 0
+          ? 'tradeChecklist.setYourOwnPrice.ourNewBtcPriceIsHigher'
+          : 'tradeChecklist.setYourOwnPrice.ourNewBtcPriceIsLower',
+        {
+          percentage: Math.abs(Number(percentageDifference)),
+        }
+      )}
+      onActionPress={() => {}}
```
kaladivo: Is empty function ok here? Maybe make the prop optional?

kaladivo (changes requested):
Good job. This is super complicated functionality and it's apparent that a lot of work went into this. 

I suggest doing a bit refactoring to separate UI from logic. Meaning, preferably moving any derivations into Jotai (try to limit usage of useCallback to derivate values).

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/PremiumOrDiscountScreen/index.tsx
SamTremko: This tempFeeAtom is just used for PremiumOrDiscountScreen, when changing the value there and on save is stored into feeAtom in state. Then in needs to be synced with state value on render so the user can see the previously set value. Maybe I can change it to useFocusEffect Maybe I don't understand what you mean by deriving so pls correct me if I am wrong.

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/FiatAmountInput.tsx
SamTremko: I had problems with debouncing this value as btcInput changes the value of fiatInput and also fiatInput changes the value of btcInput. So if fiatInput changes, then it changes the btcInput and on that change another change of fiatInput is triggered. Cause with debouncing I need to rely on some useEffect calling the calculation function. Maybe if u have some better solution I am open to implement it, maybe I just don't see it correctly.

apps/mobile/src/components/Info.tsx
```diff
@@ -8,9 +8,10 @@ import {useMemo} from 'react'
 import {atom, type PrimitiveAtom, useAtom} from 'jotai'
 
 interface Props {
-  actionButtonText: string
+  hideActionButton?: true
+  actionButtonText?: string
   text: string
-  onActionPress: () => void
+  onActionPress?: () => void
```
kaladivo, line 13: Wouldn't it make sense to display action button if onActionPress is defined? I don't like `hideActionButton?: true` definition very much

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/atoms.ts
```diff
@@ -0,0 +1,286 @@
...
+        const numberValue = Number(fiatAmount)
+        const fee = (numberValue / tradeBtcPrice) * (feeAmount / 100)
+
+        // Number cast to remove trailing zeroes after toFixed method
+        set(
+          btcValueAtom,
+          btcOrSat === 'BTC'
+            ? `${Number(
+                (numberValue / tradeBtcPrice - fee).toFixed(
+                  DECIMALS_FOR_BTC_VALUE
```
kaladivo: Is it ok to trim to 10^-6? btc, the smallest part of bitcoin is 10^-8

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/atoms.ts
```diff
@@ -0,0 +1,286 @@
...
+      fiatValueAtom: PrimitiveAtom<string>
+    }
+  ) => {
+    const tradeBtcPrice = get(tradeBtcPriceAtom)
+    const btcOrSat = get(btcOrSatAtom)
+    const feeAmount = get(feeAmountAtom)
+
+    set(fiatValueAtom, fiatAmount)
+    if (!automaticCalculationDisabled) {
+      if (tradeBtcPrice && tradeBtcPrice !== 0 && fiatAmount) {
```
kaladivo: this `Boolean(tradeBtcPrice && tradeBtcPrice !== 0)` will always evaluate the same as this `Boolean(tradeBtcPrice)`

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/atoms.ts
```diff
@@ -0,0 +1,286 @@
...
+      btcValueAtom: PrimitiveAtom<string>
+      fiatValueAtom: PrimitiveAtom<string>
+    }
+  ) => {
+    const tradeBtcPrice = get(tradeBtcPriceAtom)
+    const btcOrSat = get(btcOrSatAtom)
+    const feeAmount = get(feeAmountAtom)
+
+    set(fiatValueAtom, fiatAmount)
+    if (!automaticCalculationDisabled) {
```
kaladivo: Please use [early returns](https://gomakethings.com/the-early-return-pattern-in-javascript/) in these scenarios, it makes the code a lot more readable.

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/FiatAmountInput.tsx
```diff
@@ -0,0 +1,75 @@
...
+      onFocus={() => {
+        setIsFocused(true)
+      }}
+      onWrapperPress={() => {
+        ref.current?.focus()
+      }}
+      placeholder={`${offerForTradeChecklist?.offerInfo.publicPart.amountTopLimit}`}
+      value={fiatValue}
+      onChangeText={(input) => {
+        calculateBtcValueOnFiatAmountChange({
```
kaladivo, line 51: If I understand it correctly. If I want to fill in `0.002` value, this function will alwys trim the zeros while I type.

apps/mobile/src/utils/timedLogger.ts
```diff
@@ -1,7 +1,6 @@
 export default function timedLogger(tag: string): (s: string) => void {
   const now = Date.now()
   let latest = now
-  console.log(`Starting: ${tag}`)
```
kaladivo: Why remove this? It helps with readability. If the code, you are timing, takes long time to finish, it is useful to see, it started.

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/TradePriceTypeDialog.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const [tradePriceTypeDialogVisible, setTradePriceTypeDialogVisible] = useAtom(
+    tradePriceTypeDialogVisibleAtom
+  )
+  const freezePrice = useSetAtom(freezePriceActionAtom)
+  const setLivePrice = useSetAtom(setLivePriceActionAtom)
+
+  if (!tradePriceTypeDialogVisible) return null
+
+  return (
+    <AnimatedDialogWrapper onBackButtonPressed={() => true}>
```
kaladivo: Shouldn't onBackPress close the dialog? Just returning true here will just stop the event from bubbling

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/atoms.ts
```diff
@@ -0,0 +1,286 @@
...
+      btcValueAtom: PrimitiveAtom<string>
+      fiatValueAtom: PrimitiveAtom<string>
+    }
+  ) => {
+    const tradeBtcPrice = get(tradeBtcPriceAtom)
+    const btcOrSat = get(btcOrSatAtom)
+    const feeAmount = get(feeAmountAtom)
+
+    set(btcValueAtom, btcAmount)
+    if (!automaticCalculationDisabled) {
```
kaladivo: Early returns please. Especially when there is nested if else block.

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/atoms.ts
```diff
@@ -0,0 +1,286 @@
...
+    }
+  ) => {
+    const fiatValue = get(fiatValueAtom)
+    const btcValue = get(btcValueAtom)
+
+    set(tradeBtcPriceAtom, (prev) => (fiatValue ? Number(fiatValue) : prev))
+    set(fiatInputValueAtom, (prev) => fiatValue ?? prev)
+    set(btcInputValueAtom, (prev) => btcValue ?? prev)
+    set(tradePriceTypeAtom, 'your')
+    set(applyFeeOnTradePriceTypeChangeActionAtom)
```
kaladivo:
```suggestion
    if(fiatValue) {
      set(tradeBtcPriceAtom, Number(fiatValue))
      set(fiatInputValueAtom, fiatValue)
    }
    if(btcValue) set(btcInputValueAtom, btcValue)
    set(tradePriceTypeAtom, 'your')
    set(applyFeeOnTradePriceTypeChangeActionAtom)
```

kaladivo: Better not call `set` when value does not change

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/PremiumOrDiscountScreen/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+    useNavigation()
+
+  const offerType = useAtomValue(offerTypeAtom)
+  const feeAmount = useAtomValue(feeAmountAtom)
+  const applyFeeOnFeeChange = useSetAtom(applyFeeOnFeeChangeActionAtom)
+  const [tempFeeAmount, setTempFeeAmount] = useAtom(tempFeeAmountAtom)
+
+  useFocusEffect(
+    useCallback(() => {
+      setTempFeeAmount(feeAmount)
```
kaladivo, line 40: Might be easier and more readable to just use useState in this case

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/FiatAmountInput.tsx
```diff
@@ -0,0 +1,75 @@
...
+    >
+      <Stack>
+        <Text fos={18} ff={'$body500'} col={'$white'}>
+          {
+            currencies[
+              offerForTradeChecklist?.offerInfo.publicPart.currency ?? 'USD'
+            ].code
+          }
+        </Text>
+        {!isFocused && fiatValue && !automaticCalculationDisabled && (
```
kaladivo, line 67: Removing the label while focused is great 👍

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/CalculateAmountScreen/index.tsx
kaladivo: 👀

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/CalculateAmountScreen/components/PremiumOrDiscount.tsx
```diff
@@ -0,0 +1,70 @@
...
+    <Stack>
+      <XStack ai={'center'} jc={'space-between'} mb={'$4'}>
+        <Text fos={20} ff={'$body600'} col={'$white'}>{`% ${t(
+          'tradeChecklist.calculateAmount.premiumOrDiscount'
+        )}`}</Text>
+        <Switch
+          value={premiumOrDiscountEnabled}
+          onChange={() => {
+            setPremiumOrDiscountEnabled(!premiumOrDiscountEnabled)
+            applyFeeOnFeeChange(0)
```
kaladivo, line 36: Might be a nice touch to do this: 

```suggestion
            applyFeeOnFeeChange(premiumOrDiscountEnabled ? 0  : feeAmount )
```

kaladivo (changes requested):
Let's go through profiling today and finish the last few comments I posted so we can merge this. Great work btw, there are lot of edge cases that you had to handle and handled very well! 👍

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/PremiumOrDiscountScreen/index.tsx
kaladivo: I understand it now. This is used to set initial value on render. 

The reason why `tempFeeAmount` value is not stored in simple useState is, that it needs to be passed to `PremiumOrDiscountContent` component. Makes sense, resolving.

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/FiatAmountInput.tsx
kaladivo: I don't understand how adding debouncing would result in triggering changes that are not triggered by updates without debounce.

I am afraid, that just user typing will cause extensive rerenders but it also makes sense to update btcPriceInput when FiatInput is changed (and vice versa). So let's just make sure that calling this method does not trigger unnecessary rerenders and we're golden!

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/FiatAmountInput.tsx
kaladivo: Great work moving all callbacks to actionAtoms. It is much more readable when all updates are written in one way.

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/SetYourOwnPriceScreen/components/PriceInfo.tsx
kaladivo: ✅

kaladivo:
Please make sure to not force push into this branch 🙏. It will be easier to go through changes that way.

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/atoms.ts
```diff
@@ -169,17 +178,19 @@ export const calculateFiatValueOnBtcAmountChangeActionAtom = atom(
...
-            ? Number(btcAmount)
-            : Number(btcAmount) / SATOSHIS_IN_BTC
-        const fee = Math.round(tradeBtcPrice * numberValue) * (feeAmount / 100)
-        set(fiatValueAtom, `${Math.round(tradeBtcPrice * numberValue + fee)}`)
-      } else {
-        set(fiatValueAtom, '')
-      }
+
+    if (!btcAmount || automaticCalculationDisabled) {
+      set(fiatValueAtom, '')
```
kaladivo, line 184: Should not happen when automaticCalculationDisabled is true

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/atoms.ts
```diff
@@ -206,27 +217,28 @@ export const calculateBtcValueOnFiatAmountChangeActionAtom = atom(
...
-            : `${Math.round(
-                (numberValue / tradeBtcPrice - fee) * SATOSHIS_IN_BTC
-              )}`
-        )
-      } else {
-        set(btcValueAtom, '')
-      }
+
+    if (!fiatAmount || automaticCalculationDisabled) {
+      set(btcValueAtom, '')
```
kaladivo, line 229: Should not happen when automaticCalculationDisabled is true

kaladivo (approved):
Small comments, once fixed this is ok by me

PR #596, opened 2023-11-21
Small fixes

SamTremko:


apps/mobile/src/components/ScreenTitle/index.tsx
```diff
@@ -30,14 +30,20 @@ function ScreenTitle({
   const safeGoBack = useSafeGoBack()
   return (
     <YStack mt={'$2'} bc={'transparent'} {...props}>
-      <XStack fd="row" ai="flex-start" jc="space-between" mb="$5">
-        <Stack fs={1} maw="60%">
```
SamTremko, line 34: @kaladivo it was needed to remove the 'maw' prop from the title, as on smaller screens it can cause this weird line break (in this case word 'contacts' did not fit to one line for the 60% screen width view and current font size and font family and therefore it split weirdly into two lines)
Now we will have titles on most screens in one line (not exactly as it was in design), but we should avoid weird splitting bugs at all

apps/mobile/src/components/FaqScreen/useContent.ts
```diff
@@ -8,48 +8,91 @@ import anonymizationNoticeSvg from '../../images/anonymizationNoticeSvg'
...
+      withLink?: false
+      text: string
+      textBefore?: never
+      textAfter?: never
+      linkText?: never
+      url?: never
+    }
+  | {
+      withLink: true
+      text?: never
```
kaladivo: This line can be removed

apps/mobile/src/components/FaqScreen/useContent.ts
```diff
@@ -8,48 +8,91 @@ import anonymizationNoticeSvg from '../../images/anonymizationNoticeSvg'
...
 }
+
+type ConditionalProps =
+  | {
+      withLink?: false
+      text: string
+      textBefore?: never
+      textAfter?: never
+      linkText?: never
+      url?: never
```
kaladivo: can be removed

apps/mobile/src/components/ScreenTitle/index.tsx
kaladivo: 👍

kaladivo (changes requested):
Make sure to update all `/localization/${lang}-privacyPolicy.json` and `/localization/termsOfUse-privacyPolicy.json` in crowdin too.

kaladivo (approved):

PR #646, opened 2023-12-14
Offer expiration finetuning

SamTremko:


apps/mobile/src/components/OfferForm/components/Expiration/index.tsx
```diff
@@ -22,39 +26,85 @@ function Expiration({
   const {t} = useTranslation()
   const locale = getCurrentLocale()
   const [expirationDate, setExpirationDate] = useAtom(expirationDateAtom)
+  const [expirationDateActive, setExpirationDateActive] = useState(
+    !!expirationDate
+  )
```
kaladivo: Can this be derived from exirationDate? 

```suggestion
 const expirationDateActive === !!expirationDate
```

kaladivo (changes requested):

kaladivo (approved):

PR #648, opened 2023-12-14
Edited offer state should not be reset on re-focus

SamTremko:


apps/mobile/src/components/ModifyOffer/components/EditOfferScreen.tsx
```diff
@@ -32,27 +31,22 @@ type Props = RootStackScreenProps<'EditOffer'>
 
 function EditOfferScreen({
   route: {
-    params: {offerId},
+    params: {offer},
```
kaladivo: https://reactnavigation.org/docs/params/#what-should-be-in-params

> Avoid passing the full data which will be displayed on the screen itself (e.g. pass a user id instead of user object).

apps/mobile/src/components/OfferDetailScreen/index.tsx
```diff
@@ -1,52 +1,20 @@
...
-import {isSome} from 'fp-ts/Option'
-import Button from '../Button'
 import KeyboardAvoidingView from '../KeyboardAvoidingView'
 
 type Props = RootStackScreenProps<'OfferDetail'>
 
 function OfferDetailScreen({
   route: {
-    params: {offerId},
+    params: {offer},
```
kaladivo: https://reactnavigation.org/docs/params/#what-should-be-in-params

> Avoid passing the full data which will be displayed on the screen itself (e.g. pass a user id instead of user object).

kaladivo (changes requested):

SamTremko:
@kaladivo I amended the commit, as I quite completely removed changes from previous one and did the logic in other way

kaladivo (approved):

PR #661, opened 2024-01-04
release 1.12.0

kaladivo:


kaladivo:
# Chat compatibility test cases
There were a lot of changes in chat messages processing. We need to ensure we don't introduce any bugs. Please find test cases below: 

## Functionality that was not changed
- Download any version `1.11.x` and lower
- Login
- Send and receive every message that is possible to receive - you can do that by performing actions listed: 
  - Send a text message
  - Send a text message as a quote response
  - Reveal identity
  - Contact reveal
  - Close chat
  - Deny messaging request
- Update the app to `1.12.0`
- All messages should still be visible
- You should be able to send all messages types (see list above)

## New functionality - Incompatible messages
- Make sure you have version `1.12.0` 
- Close the app on your device 
- Let @kaladivo know, he will send you a message that is meant for newer Vexl version
- You should receive push notification with general new message text
- Open the chat
- You should see the message as not compatible with your version of Vexl
- On chat list you should see Incompatible messages preview.
- Update to version 1.12.1
- You should see the message decoded with proper contents

PR #771, opened 2024-01-30
Add option to change fiat currency in trade checklist

SamTremko:


SamTremko:
Current dependencies on/for this PR:
* `main`
  * **PR #771** <a href="https://app.graphite.dev/github/pr/vexl-it/vexl/771?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/vexl-it/vexl/771?utm_source=stack-comment).

apps/mobile/src/components/ChatDetailScreen/atoms/index.tsx
```diff
@@ -778,6 +780,23 @@ export const chatMolecule = molecule((getMolecule, getScope) => {
...
+    return (
+      tradeChecklistAmountData?.amountData.currency ??
+      originOffer?.offerInfo?.publicPart?.currency ??
+      'USD'
+    )
+  })
+
+  const btcPriceForTradeCurrencyAtom = atom((get) => {
+    return get(createBtcPriceForCurrencyAtom(tradeOrOriginOfferCurrencyAtom))
+  })
```
kaladivo: When possible, avoid creating atom inside atom functions

```suggestion
 const btcPriceForTradeCurrencyAtom = createBtcPriceForCurrencyAtom(tradeOrOriginOfferCurrencyAtom)
```

apps/mobile/src/components/ChatDetailScreen/components/VexlbotMessageItem/components/TradeChecklistAmountView.tsx
```diff
@@ -69,12 +69,6 @@ function TradeChecklistAmountView(): JSX.Element | null {
     submitTradeChecklistUpdates,
   ])
 
-  useFocusEffect(
-    useCallback(() => {
-      void refreshBtcPrice(offerCurrency)()
-    }, [refreshBtcPrice, offerCurrency])
-  )
-
```
kaladivo, line 77: Does this need to be removed? The price fetch needs to be triggered for it to be in state.

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/atoms.ts
```diff
@@ -26,12 +30,26 @@ export const tradeBtcPriceAtom = atom<number>(0)
...
+export const btcPriceForOfferWithStateAtom = atom((get) => {
+  const selectedCurrencyCode = get(selectedCurrencyCodeAtom)
+  const tradeOrOriginOfferCurrency = get(tradeOrOriginOfferCurrencyAtom)
+
+  return get(
+    createBtcPriceForCurrencyAtom(
+      selectedCurrencyCode ?? tradeOrOriginOfferCurrency ?? 'USD'
+    )
+  )
+})
```
kaladivo:
```suggestion typescript
export const currencyToUseAtom = atom(get(): CurrencyCode => {
  return get(selectedCurrencyCodeAtom) ?? get(tradeOrOriginOfferCurrencyAtom) ?? 'USD'
})

export const btcPriceForOfferWithStateAtom = createBtcPriceForCurrencyAtom(currencyToUseAtom)
```

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/atoms.ts
```diff
@@ -230,6 +242,30 @@ export const calculateBtcValueOnFiatAmountChangeActionAtom = atom(
   }
 )
 
+export const refreshCurrentBtcPriceActionAtom = atom(null, (get, set) => {
+  const currency =
+    get(selectedCurrencyCodeAtom) ??
+    get(tradeOrOriginOfferCurrencyAtom) ??
+    'USD'
```
kaladivo:
```suggestion
get(currencyToUseAtom)
```

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/atoms.ts
```diff
@@ -230,6 +242,30 @@ export const calculateBtcValueOnFiatAmountChangeActionAtom = atom(
...
+
+  return pipe(
+    set(refreshBtcPriceActionAtom, currency),
+    T.map(() => {
+      set(
+        tradeBtcPriceAtom,
+        (prev) => get(btcPriceForOfferWithStateAtom)?.btcPrice ?? prev
+      )
+      // we need to recalculate amount of fiat based on new btc price
+      set(calculateFiatValueOnBtcAmountChangeActionAtom, {
```
kaladivo, line 256: To me this seems like an antipattern. 

Atoms are reactive and they should react on other atoms' changes.

Calling set atom with values that were retrieved from other atoms is not the right approach (for example `btcAmount` parameter). 

The proper way how to do this would be to move this logic into atom's getter and just set atom value here.

I understand that it would be pretty significant logic change to fix this here, but it adds unnecessary complexity that will cause small changes being harder and harder to implement. 

Please add todo and we'll fix this next time when (if) we revisit this feature and have more time...

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/atoms.ts
```diff
@@ -245,6 +281,28 @@ export const toggleBtcOrSatValueActionAtom = atom(
...
+export const toggleFiatCurrencyActionAtom = atom(
+  null,
+  (get, set, currency: CurrencyCode) => {
+    const btcInputValue = get(btcInputValueAtom)
+    const btcPriceData = get(btcPriceDataAtom)[currency]
+
+    set(selectedCurrencyCodeAtom, currency)
+
+    if (!btcPriceData || btcPriceData.state === 'error') {
+      void set(refreshCurrentBtcPriceActionAtom)()
```
kaladivo: Why is this called only on error? Why not call `set(refreshCurrentBtcPriceActionAtom)()` every time and let the caching take care of not sending too many requests?

Even the logics in these atoms are the same.

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/CalculateAmountScreen/index.tsx
```diff
@@ -86,16 +88,19 @@ function CalculateAmountScreen({
         void submitTradeChecklistUpdates()().finally(() => {
           showLoadingOverlay(false)
         })
+        navigation.navigate('ChatDetail', store.get(chatWithMessagesKeys))
+      } else {
+        goBack()
```
kaladivo: Is this correct? Can `navigation.navigate('AgreeOnTradeDetails')` be called instead?

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/FiatAmountInput.tsx
```diff
@@ -1,37 +1,85 @@
...
 interface Props {
   automaticCalculationDisabled?: boolean
   fiatValueAtom: PrimitiveAtom<string>
   btcValueAtom: PrimitiveAtom<string>
   showSubtitle?: boolean
+  editable?: boolean
 }
 
+const fiatCurrenciesDropdownData: Array<DropdownItemProps<CurrencyCode>> =
+  Object.values(currencies).map((currency) => ({
```
kaladivo: Why are we iterating over localization currencies here? 

Woudn't it make more sense to iterate over `AcceptedCurrency`  (from packages/rest-api/src/services/btcPrice.ts) - for those we can actually fetch btcPrice  or `CurrencyCode` which is accepted by `toggleFiatCurrencyActionAtom`?

kaladivo (changes requested):
How is the price refreshed. I found 2 places 1. syncDataWithChatState - that one is called at the begining, 2. when toggling fiat currency. 

Once the price for currency is fetched it won't be removed and refresh should be called to ensure the price is updated.

kaladivo:
After further investigation, it looks like price is requested after every screen open - `syncDataWithChatState` call is sufficient to refresh the price. 

Removing comments regarding that

apps/mobile/src/components/ChatDetailScreen/components/VexlbotMessageItem/components/TradeChecklistAmountView.tsx
kaladivo: `syncDataWithChatState` call is sufficient to refresh the price.

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/atoms.ts
kaladivo: I get that now. We don't expect to spend enough time on this screen to make the price significally outdated. 

But I think calling just `void set(refreshCurrentBtcPriceActionAtom)()` (removing the condition and the else if branch) won't change the functionality, right?

SamTremko: Yes it will not change the functionality, this fetch is needed as the user can change currency to some, that has not been fetched before. So calling just refresh without conditions is OK I think

kaladivo: Let's do it then. It will simplify it at least a little bit

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/atoms.ts
SamTremko: Yeah you are right, I was also thinking the same when writing this code, but did not want to change this calculate function to be able to deliver the feature, but I totally agree with you

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/FiatAmountInput.tsx
SamTremko: Good point, those two can differ

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/atoms.ts
kaladivo: Let's add the todo comment at least

apps/mobile/src/components/ChatDetailScreen/atoms/index.tsx
SamTremko: ou sh*it thx for noticing!

apps/mobile/src/components/TradeChecklistFlow/components/CalculateAmountFlow/components/FiatAmountInput.tsx
```diff
@@ -1,37 +1,86 @@
...
 interface Props {
   automaticCalculationDisabled?: boolean
   fiatValueAtom: PrimitiveAtom<string>
   btcValueAtom: PrimitiveAtom<string>
   showSubtitle?: boolean
+  editable?: boolean
 }
 
+const fiatCurrenciesDropdownData: Array<DropdownItemProps<CurrencyCode>> =
+  Object.values(AcceptedCurrency).map((currency: AcceptedCurrency) => ({
```
kaladivo: `Object.values` call is redundant
```suggestion
  AcceptedCurrency.map((currency: AcceptedCurrency) => ({
```

kaladivo (approved):
Small changes

SamTremko:
### Merge activity

* **Jan 30, 3:20 PM**: @SamTremko merged this pull request with [Graphite](https://app.graphite.dev/github/pr/vexl-it/vexl/771).

PR #782, opened 2024-01-31
Add phone number bug in chat

SamTremko:


SamTremko:
Current dependencies on/for this PR:
* `main`
  * **PR #782** <a href="https://app.graphite.dev/github/pr/vexl-it/vexl/782?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #783** <a href="https://app.graphite.dev/github/pr/vexl-it/vexl/783?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/vexl-it/vexl/782?utm_source=stack-comment).

apps/mobile/src/components/ChatDetailScreen/components/QuickActionBanner.tsx
```diff
@@ -324,14 +329,12 @@ function QuickActionBanner(): JSX.Element | null {
...
-          const fullPhoneNumber =
-            otherSideData.fullPhoneNumber ??
-            lastMessage.message.deanonymizedUser?.fullPhoneNumber
+          const fullPhoneNumber = otherSideData.fullPhoneNumber
 
           if (fullPhoneNumber) {
             void addRevealedContact({
               name: fullPhoneNumber,
-              normalizedNumber: E164PhoneNumber.parse(fullPhoneNumber),
+              normalizedNumber: fullPhoneNumber,
```
kaladivo, line 331: Good catch, no need to parse the number again

apps/mobile/src/state/chat/atoms/revealContactActionAtom.ts
```diff
@@ -113,17 +114,30 @@ export default function revealContactActionAtom(
...
         TE.map(() => {
           const successMessage: ChatMessageWithState = {
-            message: chatMessage,
+            message,
             state: 'sent',
           }
+
+          const contactRevealMessage = get(chatWithMessagesAtom).messages.find(
+            (one) => one.message.messageType === 'REQUEST_CONTACT_REVEAL'
+          )
```
kaladivo: Can we do this inside the condition below?

apps/mobile/src/state/chat/utils/addRealLifeInfoToChat.ts
```diff
@@ -27,19 +27,21 @@ export default function addRealLifeInfoToChat(
...
+    }
+
+    if (contactRevealData) {
+      const dataFromContactReveal =
+        processTradeChecklistContactRevealMessageIfAny(
+          contactRevealData,
+          chat.chat.otherSide.realLifeInfo
+        )
       return setRealLifeInfo(dataFromContactReveal)(chat)
+    }
```
kaladivo: What if identityRevealData and contactRevealData are both nonnull? Only identityRevealData will be processed in that case

packages/domain/src/general/DeanonymizedUser.ts
```diff
@@ -1,9 +1,10 @@
 import {z} from 'zod'
+import {E164PhoneNumber} from './E164PhoneNumber.brand'
 import {UserName} from './UserName.brand'
 
 export const DeanonymizedUser = z.object({
   name: UserName,
   partialPhoneNumber: z.string().optional(),
-  fullPhoneNumber: z.string().optional(),
+  fullPhoneNumber: E164PhoneNumber.optional(),
```
kaladivo, line 8: 👍

apps/mobile/src/components/ChatDetailScreen/components/QuickActionBanner.tsx
```diff
@@ -311,7 +310,13 @@ function QuickActionBanner(): JSX.Element | null {
     )
   }
 
-  if (contactRevealStatus === 'shared') {
+  if (
+    (lastMessage.message.messageType === 'APPROVE_CONTACT_REVEAL' ||
+      (lastMessage.message.messageType === 'TRADE_CHECKLIST_UPDATE' &&
+        lastMessage.message.tradeChecklistUpdate?.contact?.status ===
+          'APPROVE_REVEAL')) &&
```
kaladivo: We want to show this only if the last message is response on reveal? Is that desired? If I approve contact reveal and I send text  message right after that, will the other side see the banner? Should it?

kaladivo (changes requested):

apps/mobile/src/state/chat/utils/addRealLifeInfoToChat.ts
SamTremko: yes that is true, but this should not happen, as we are using this function in per message processing

kaladivo: Can we split it into two? Otherwise I guaratee you, in the future, one of us will call it, thinking it does both.

kaladivo (approved):

kaladivo (changes requested):
Jus one more comment

kaladivo (approved):

PR #1048, opened 2024-05-07
chore: Update README.md build instructions

TomKalina:
Change to the project startup instructions. 

The build line no longer exists
```
cd apps/mobile
```

The line to run is platform dependent, doesn't make sense for Android
```
npx expo run:ios
```


kaladivo:
`expo start` will start expo dev server but it won't install the dev build on a device. That's what the `expo run:ios` / `expo run:android` is for. 

I'll update the readme accordingly to include the latest build instructions (#1119)

PR #1159, opened 2024-06-18
Share information about interaction with offer as snackbar

SamTremko:


apps/mobile/src/components/ToastNotification/index.tsx
```diff
@@ -31,8 +33,8 @@ function ToastNotification(): JSX.Element {
 
   return (
     // this view has to be present in tree
-    // without it existing animation on Toast will not work
-    <Stack pos="absolute" alignSelf="center" top={-10}>
```
kaladivo, line 35: Let's not change this. But I would prefer to not use the shortcuts. The code is more readable that way. But it's up for the discusion

apps/mobile/src/components/ChatDetailScreen/components/RequestScreen.tsx
```diff
@@ -76,6 +79,38 @@ function RequestScreen(): JSX.Element {
...
+    if (hasPreviousCommunication === 'anotherInteractionWithHistory') {
+      return t('messages.youHaveAlreadyInteractedWithThisUser')
+    }
+
+    return undefined
+  }, [hasPreviousCommunication, requestedByMe, t])
+
+  useEffect(() => {
+    if (previousCommunicationInfoMessage) {
+      setToastNotification({
```
kaladivo, line 100: That is cool!

kaladivo (approved):

PR #1165, opened 2024-06-25
Fix toast notification layout

SamTremko:


kaladivo (approved):
https://github.com/Effect-TS/effect/blob/main/packages/schema/README.md#introduction

PR #1178, opened 2024-07-16
Final touches for progressive offer creation

SamTremko:


apps/mobile/src/components/CRUDOfferFlow/atoms/offerFormStateAtoms.ts
```diff
@@ -1110,16 +1126,31 @@ export const offerFormMolecule = molecule(() => {
...
+    if (reason === 'noLocationForInPersonOffer')
+      Alert.alert(t('offerForm.errorLocationNotFilled'))
+    if (reason === 'priceNotFilled')
+      Alert.alert(t('offerForm.errorPriceNotFilled'))
+    if (reason === 'deliveryMethodNotFilled')
+      Alert.alert(t('offerForm.errorDeliveryMethodNotFilled'))
+    if (reason === 'pickupLocationNotFilled')
+      Alert.alert(t('offerForm.errorPickupLocationNotFilled'))
+    if (reason === 'exceededLimit')
+      Alert.alert(t('offerForm.errorExceededLimits'))
```
kaladivo: Theoretically, you can return the offer form string right away from the `dontAllowNavigationToNextStepAndReturnReasonAtom` atom. So you don't have to do this, but just 

```
Alert.alert(`offerFor.${reason}`)
```

It just needs to be returned `as const` to make the typechecking happy.

apps/mobile/src/components/CRUDOfferFlow/index.tsx
```diff
@@ -85,18 +96,25 @@ function CRUDOfferFlow({route: {params}, navigation}: Props): JSX.Element {
...
       const currentPage =
-        screensBasedOnListingType[pageIndex] ?? 'ListingAndOfferType'
-      setCurrentStepInOfferCreation(currentPage)
-      navigation.navigate('CRUDOfferFlow', {
-        screen: currentPage,
-        offerId: params.offerId,
-      })
-      setPage(pageIndex)
+        screensBasedOnListingType[prevOrNextPageIndex] ?? 'ListingAndOfferType'
+      if (prevOrNextPageIndex < page || !dontAllowNavigationToNextStep) {
```
kaladivo: IF you only need dontAllowNavigationToNextStep anywhere else, you can do: 

```suggestion
      if (prevOrNextPageIndex < page || !store.get(dontAllowNavigationToNextStepAndReturnReasonAtom)) {
```

kaladivo (approved):

PR #1190, opened 2024-07-21
Fee should be applied on fiat and not on BTC

SamTremko:


apps/mobile/src/components/ChatDetailScreen/components/VexlbotMessageItem/components/TradeChecklistAmountView.tsx
```diff
@@ -76,19 +75,19 @@ function TradeChecklistAmountView(): JSX.Element | null {
...
   )
 
-  const btcAmount = useMemo(
+  const fiatAmount = useMemo(
     () =>
-      amountDataToDisplay?.amountData.btcAmount
-        ? formatBtcPrice(
-            applyFeeOnBtcAmount(
-              amountDataToDisplay.amountData.btcAmount,
+      amountDataToDisplay?.amountData.fiatAmount
```
kaladivo: So the value in amountDataToDispaly.amountData.fiatAmount is not actually the amount that is displayed right? Would it be possible to move the memo logic inside the atom?

apps/mobile/src/components/TradeCalculator/atoms.ts
```diff
@@ -65,47 +52,49 @@ export const btcPriceForOfferWithStateAtom =
...
+
+  set(tradePriceTypeAtom, 'your')
+  // set(applyFeeOnTradePriceTypeChangeActionAtom)
+})
+
 export const applyFeeOnTradePriceTypeChangeActionAtom = atom(
   null,
   (get, set) => {
-    const btcInputValue = Number(get(btcInputValueAtom))
+    const fiatInputValue = Number(get(fiatInputValueAtom))
```
kaladivo, line 91: Can it happen that the number cast fails? If so what will happen in the app?

kaladivo (approved):

kaladivo (approved):

apps/mobile/src/components/TradeCalculator/atoms.ts
SamTremko: I checked it, this should be OK, as input value is checked for non numeric chars on each change and in the case of occurance is the value of the input replaced with zero string, like '0'

PR #1206, opened 2024-07-29
Separate BTC exchange rate service from location service

SamTremko:


apps/btc-exchange-rate-service/.env.example
kaladivo, line 1: delete or copy over

apps/btc-exchange-rate-service/.env.test
kaladivo, line 1: remove non needed envs

kaladivo (approved):

PR #1305, opened 2024-10-08
fix: user-contact table should have unique hash_from,hash_to

kaladivo:


SamTremko (approved):

kaladivo:
```sql
select count(*), array_agg(id)
from user_contact
group by hash_from, hash_to
having count(*) > 5
```

```sql
select count(*) from (select count(*), array_agg(id)
from user_contact
group by hash_from, hash_to
having count(*) > 1)as a
```
currently to 15.10. 11:16 returns: 
155 655

kaladivo:
The number seems to be decreased now. We are at: 154 118.

PR #1325, opened 2024-10-17
feat: add more offers metrics

kaladivo:


apps/offer-service/src/metrics.ts
```diff
@@ -1,15 +1,24 @@
...
+import {Array, Effect, Layer, pipe} from 'effect'
 
 const OFFER_PUBLIC_PART_DELETED = 'OFFER_PUBLIC_PART_DELETED' as const
 const OFFER_MODIFIED = 'OFFER_MODIFIED' as const
 const OFFER_CREATED = 'OFFER_CREATED' as const
 const OFFER_REPORTED = 'OFFER_REPORTED' as const
+const TOTAL_BUY_OFFERS = 'TOTAL_BUY_OFFERS' as const
+const TOTAL_SELL_OFFERS = 'TOTAL_SELL_OFFERS' as const
+const TOTAL_SELL_OFFERS_EXPIRED = 'TOTAL_SELL_OFFERS' as const
+const TOTAL_BUY_OFFERS_EXPIRED = 'TOTAL_SELL_OFFERS' as const
```
SamTremko: I think values here should be:
`const TOTAL_SELL_OFFERS_EXPIRED = 'TOTAL_SELL_OFFERS_EXPIRED' as const
const TOTAL_BUY_OFFERS_EXPIRED = 'TOTAL_BUY_OFFERS_EXPIRED' as const` ?

SamTremko (approved):

apps/offer-service/src/metrics.ts
kaladivo: You are right, thanks for the check!

PR #1342, opened 2024-10-23
fix-1327: Fix issue with checklist not reacting to user input

SamTremko:


apps/mobile/src/components/ChatDetailScreen/components/VexlbotMessageItem/components/TradeChecklistMessageWrapper.tsx
```diff
@@ -0,0 +1,28 @@
...
+  const {chatIdAtom, publicKeyPemBase64Atom} = useMolecule(chatMolecule)
+
+  const chatId = useAtomValue(chatIdAtom)
+  const inboxKey = useAtomValue(publicKeyPemBase64Atom)
+  const setParentChat = useSetAtom(fromChatAtoms.setParentChatActionAtom)
+
+  // sets parent chat for checklist as user can interact with it directly from chat
+  // without going through trade checklist flow
+  useEffect(() => {
+    setParentChat({chatId, inboxKey})
```
kaladivo: Shouldn't this be done once for the chat as a whole? It will be called many times unnecessarily for every message instead of once when the chat is rendered...

kaladivo (changes requested):
I would rather make sure that the chat is set for the trading checklist at the chat root then on every message individually

apps/mobile/src/components/ChatDetailScreen/components/VexlbotMessageItem/components/TradeChecklistMessageWrapper.tsx
SamTremko: Previously I was also thinking about that, but then I wanted to split the checklist functionality from chat, so chat would not set the checklist state, if you know what I mean. Thats why I did it this way.

SamTremko: But as I am thinking about it, you are right, as the checklist is part of chat schema, it is either way connected with it. Will rework it, good point.

kaladivo (approved):

PR #1351, opened 2024-10-31
fix: metrics key should be configurable so it is different on prod and stage

kaladivo:


kaladivo:
Was hotfix already released on prod

SamTremko (approved):

Issue #1354, opened 2024-11-01
batch uploading contacts


kaladivo:
1. Upload contacts in batches
2. When adding single contact just add one contact dont sent all selected contacts again
3. When removing contact, remove just the one contact

Other things: 
1. make sure quotas still work
2. Create new endpoint to add / delete contact for user
3. ‼️ Handle new user notification properly - this is the most complex task about this issue

PR #1371, opened 2024-11-07
feat: #1354 - batch uploading contacts

kaladivo:


apps/mobile/src/state/contacts/atom/submitContactsActionAtom.ts
```diff
@@ -44,90 +50,142 @@ export const submitContactsActionAtom = atom(
...
+                  contacts: chunkToImport.map((one) => one.computedValues.hash),
+                  replace,
+                },
+              })
+              .pipe(
+                Effect.zipLeft(
+                  Ref.update(
+                    importedNumbersSoFarRef,
+                    HashSet.union(
+                      newContactsToImport.map(
```
SamTremko: Shouldn't be here `chunkToImport` instead of `newContactsToImport`? Or maybe I don't understand it correctly ...

SamTremko (approved):

apps/mobile/src/state/contacts/atom/submitContactsActionAtom.ts
kaladivo: You are right! Good catch!

Issue #679, opened 2024-01-18
Delete deleted chats from filesystem


kaladivo:
When chat is deleted. Remove the chats from mmkv storage & remove all resources (images) created by chat.

kaladivo:
Delete data for all chats that were closed more than 30 days ago

kaladivo:
Check in hook when the app is launched

PR #1372, opened 2024-11-07
feat: #679 - Delete deleted chats from file system

SamTremko:


SamTremko:
@kaladivo I did it a little differently as a part of fetching messages for all inboxes, we can discuss this

apps/mobile/src/state/chat/atoms/checkAndDeleteOldChatsDataActionAtom.ts
```diff
@@ -0,0 +1,42 @@
...
+      ...prev,
+      chats: prev.chats.filter((oneChat) => {
+        const isChatDeletedAndExpired = oneChat.messages.some(
+          (message) =>
+            (message.message.messageType === 'DELETE_CHAT' ||
+              message.message.messageType === 'INBOX_DELETED' ||
+              message.message.messageType === 'CANCEL_REQUEST_MESSAGING' ||
+              message.message.messageType === 'BLOCK_CHAT' ||
+              message.message.messageType === 'DISAPPROVE_MESSAGING' ||
+              message.message.messageType === 'OFFER_DELETED') &&
```
kaladivo: Why not use already existing function for this. Such as: 

https://github.com/vexl-it/vexl/blob/c388e104d7feaec2dc3eddc28d0f90f92a625bf4/apps/mobile/src/state/chat/utils/isChatActive.ts

kaladivo: The problem is, that INBOX_DELETED might be in the chat sent by the other side, in that case, we don't want to delete the chat before user deletes it first by clicking on "delete chat" button

apps/mobile/src/state/chat/atoms/checkAndDeleteOldChatsDataActionAtom.ts
```diff
@@ -0,0 +1,42 @@
+import {type PublicKeyPemBase64} from '@vexl-next/cryptography/src/KeyHolder'
+import {atom} from 'jotai'
+import {focusAtom} from 'jotai-optics'
+import {DateTime} from 'luxon'
+import removeFile from '../../../utils/removeFile'
+import messagingStateAtom from './messagingStateAtom'
+
+const ONE_MONTH_IN_MS = 1000 * 60 * 60 * 24 * 30
+
+export const checkAndDeleteOldChatsDataActionAtom = atom(
```
kaladivo: Also would it be possible to remove the inbox also? (only if there is no offer for it nor there is any other open chats)

apps/mobile/src/state/marketplace/index.ts
```diff
@@ -540,6 +541,25 @@ export const deleteOffersActionAtom = atom<
               !adminIdsToDelete.includes(o.ownershipInfo?.adminId)
           )
         )
+
+        // Only when deleting one offer
+        if (adminIdsToDelete.length === 1) {
```
kaladivo: This will run only when deleting offer. It might be good idea to iterate through all inboxes periodically and check wether they should be removed or not (if there are none offers nor chats)

kaladivo:
Sorry, I updated the branch by mistake :/. Clicked on wrong pull request. But nothing was lost

apps/mobile/src/state/chat/atoms/checkAndDeleteEmptyInboxesWithoutOfferAtom.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const checkAndDeleteEmptyInboxesWithoutOfferAtom = atom(
+  null,
+  (get, set) => {
+    const myOffersIds = get(myOffersAtom).map(
+      (offer) => offer.offerInfo.offerId
+    )
+    set(
+      messagingStateAtom,
+      flow(
```
kaladivo: Unnecessary flow

apps/mobile/src/state/chat/atoms/checkAndDeleteEmptyInboxesWithoutOfferAtom.ts
```diff
@@ -0,0 +1,27 @@
...
+        Array.filter(
+          (one) =>
+            // is my inbox
+            !one.inbox.offerId ||
```
kaladivo: Can we check for my inbox by comparing public key to session public key? I would like to prevent bugs if we decide in the future, that inboxes without offerId can be created....

kaladivo (approved):

PR #1389, opened 2024-11-19
fix: contacts being erased when incrementally updating

kaladivo:


apps/mobile/src/state/contacts/atom/submitContactsActionAtom.ts
```diff
@@ -78,22 +78,35 @@ export const submitContactsActionAtom = atom(
...
+
+      const newContactsToImport = doIncrementalUpdate
+        ? contactsThatShouldBeImported.filter((one) => !one.flags.imported)
+        : contactsThatShouldBeImported
 
       const importedNumbersSoFarRef = yield* _(
         Ref.make(HashSet.empty<E164PhoneNumber>())
       )
+
+      if (newContactsToImport.length === 0 && !doIncrementalUpdate) {
```
SamTremko, line 93: I am not sure if I understand it correctly, but in situation, when I have no new contacts to import and I want just remove 1 number from already imported contacts, this would be true?

SamTremko (approved):

PR #1411, opened 2024-12-09
Revert: expo upgrade

kaladivo:


kaladivo:
Not needed anymore see #1412

PR #1456, opened 2025-01-18
fix #1431: Check app UX for increased system font size

### LINKED ISSUE #1431: Check app UX when user has increased the system font size
Snímka obrazovky 2024-12-27 o 8 40 00

SamTremko:


yarn.lock
```diff
@@ -25617,6 +25618,15 @@ __metadata:
   languageName: node
   linkType: hard
 
+"react-native-device-info@npm:^14.0.2":
```
kaladivo, line 25621: Does this library add additional permissions to Android app? Can you please check? We don't want to have something like `add id` permission required because we install this library

kaladivo (changes requested):

yarn.lock
SamTremko: @kaladivo I checked the manifest, and no other permissions were added there. From the documentation point of view, we would need to add some permissions once we want to use some interfaces provided by this library (MAC address, IP address or so ...), but none of those we need for the use case of font size

kaladivo (approved):

Issue #1531, opened 2025-02-04
Too much information in the location section


stepanborovka:
Can we make something with the location section? Right now it is showing so much information, that it does not look good and the offers are so huge right now. What about narrowing it down? It would make my experience much smoother.

![Image](https://github.com/user-attachments/assets/2449d7ea-92de-4276-bd38-bc0b4c4d5274)

leomira:
I totally agree.

I propose: 
- get rid of the information about the range (you can see it on the map)
- get rid of the country information, keep only the city + city part

stepanborovka:
Listing of country can be useful, but I know we have map so no point for that in the description.

leomira:
@stepanborovka yeah, it might be - I would give it a try without and we can bring it back if it was too confusing for users.

kaladivo:
Should go like: 
City district (if visible) - city. 

(later we add country code)

PR #1533, opened 2025-02-10
feat: #1531 - shorten location text

kaladivo:


packages/rest-api/src/services/location/specification.ts
```diff
@@ -29,7 +29,7 @@ export const GetGeocodedCoordinatesEndpoint = Api.get(
 ).pipe(
   Api.setRequestQuery(GetGeocodedCoordinatesRequest),
   Api.setResponseBody(GetGeocodedCoordinatesResponse),
-  Api.setSecurity(ServerSecurity),
+  // Api.setSecurity(ServerSecurity),
```
SamTremko: Why are we removing security here?

SamTremko (approved):

Issue #1500, opened 2025-01-27
Clubs BE - modify offer service to give offers to different than master public key


kaladivo:
Currently the offers are returned for a public key in auth headers. For club implementation to work we need a way so user can query offers for different public key. 

AC: 
- add endpoint to fetch new offers where public key can be specified. User should sign server secret to prove he really owns the public key (similarly to chat service).

PR #1537, opened 2025-02-11
feat #1500: Modify offer service to be used with clubs

SamTremko:


kaladivo (approved):

packages/server-utils/src/services/challenge/db/ChallegeDbService/configs.ts
```diff
@@ -0,0 +1,5 @@
+import {Config} from 'effect'
+
+export const challengeExpirationMinutesConfig = Config.number(
+  'CHALLENGE_EXPIRATION_MINUTES'
```
kaladivo, line 4: add default value to be 10

Issue #1628, opened 2025-04-09
Add blog CRM


leomira:

kaladivo:
Where should the blog button be? @leomira

leomira:
@kaladivo marketplace, next to trade calculator and meetups & clubs

kaladivo:
Release as hotifx

PR #1668, opened 2025-04-24
feat: #1628 - add blog into app

kaladivo:


apps/content-service/src/handlers/events.ts
```diff
@@ -0,0 +1,152 @@
...
+import {Handler} from 'effect-http'
+import {clearCacheTokenHashConfig} from '../configs'
+import {WebflowCmsService} from '../utils/webflowCms'
+import {
+  type WebflowEventItem,
+  type WebflowSpeakerItem,
+} from '../utils/webflowCms/domain'
+
+const EVENTS_REDIS_KEY = 'CONTENT:events'
+const EVENTS_REDIS_LIVETIME_MILISEC = 1000 * 60 * 60 * 24
```
SamTremko: Maybe these keys can be imported from:
`apps/content-service/src/utils/cache.ts`
Or do we need them twice?

SamTremko (approved):

PR #1703, opened 2025-05-09
fix: Issues with club reporting and making club inactive

SamTremko:


apps/mobile/src/components/OfferDetailScreen/atoms.ts
```diff
@@ -98,10 +98,13 @@ export const reportOfferActionAtom = atom(
               Array.contains(clubUuid)(offer.offerInfo.privatePart.clubIds)
           ),
           Option.map(([_, keyPair]) =>
-            api.offer.reportClubOffer({
-              offerId: offer.offerInfo.offerId,
-              keyPair,
-            })
+            pipe(
```
kaladivo: Run in yield `effect.flatten`and make sure to handle errors `effect.ignore` or similar...

kaladivo (approved):

PR #1829, opened 2025-06-27
fix #1827: Unable to change profile picture

### LINKED ISSUE #1827: Bug: unable to select profile picture

SamTremko:


apps/mobile/src/utils/imagePickers.ts
```diff
@@ -46,7 +45,7 @@ export function moveImageToInternalDirectory({
       if (!rootDirectory) throw new Error('document dir not found')
 
       const parentDirectory = directory
-        ? urlJoin(rootDirectory, directory)
+        ? `${rootDirectory}${directory}`
```
kaladivo, line 48: Really? This seems weird. I would rather use function for joining urls that concating it directly like this

kaladivo (changes requested):

kaladivo (approved):

PR #1978, opened 2025-10-01
Android: Add Adaptive Icons support

dennisreimann:
Adds [Adaptive Icons](https://developer.android.com/develop/ui/views/launch/icon_design_adaptive) support on Android.

Unfortunately, I wasn't able to test this locally due to build problems, but according to the [Expo icon docs](https://docs.expo.dev/develop/user-interface/splash-screen-and-app-icon/#android) this should work.


kaladivo:
Added in: https://github.com/vexl-it/vexl/pull/2299. Will go out in next release :)

PR #1929, opened 2025-09-11
feat #1927: Expo SDK 54 update

SamTremko:

PR #1989, opened 2025-10-10
Fix multiple small fixes

SamTremko:


SamTremko:
Let's merge [Expo SDK 54 update](https://github.com/vexl-it/vexl/pull/1929) first, as this one is created on top of that branch

SamTremko:
Not relevant, will be new one

PR #1992, opened 2025-10-15
fix #1981: Update error message for SMS verification

### LINKED ISSUE #1981: Update error message for SMS verification
When user has a VPN on, and we don't let them through with SMS verification, please update the error message and recommend turning off VPN + some message like: Didn't work? Reach out to our support. 

Thx!

SamTremko:


kaladivo (approved):
Nice work

PR #2035, opened 2025-11-12
chore: Update patch versions of tamagui and expo packages

SamTremko:


kaladivo (approved):
Just FYI, there is not patch update.

PR #2112, opened 2025-12-18
fix #1951: Check redis lock times and increase for attempts where loc…

### LINKED ISSUE #1951: Check lock times: Attempted to release an expired lock.

SamTremko:
…k expired before release


apps/contact-service/src/utils/NewClubUserNotificationService.ts
```diff
@@ -182,7 +182,8 @@ export class NewClubUserNotificationsService extends Context.Tag(
               Effect.all(
                 Array.map(notificationsToSend, (one) =>
                   sendExpoNotificationToAllHandleNonExistingTokens(one)
-                )
+                ),
+                {concurrency: 10}
```
SamTremko: and also here

kaladivo (approved):

Issue #1003, opened 2024-04-24
Remove no longer used keys from crowdin


SamTremko:
- filterOffers.searchByText

PR #2113, opened 2025-12-19
chore #1003: Add script to remove no longer used keys from crowdin

SamTremko:


kaladivo (approved):
Nice work with the claude :)

PR #2332, opened 2026-04-03
This is adding: 'More settings' menu + toggle for: .vcf import/export, store in phone contacts, delete Vexl only contact.

KusAdama:
There are a few changes I'm proposing, packed them together because it's a fit for what is done.

Before explaining point by point, let me address one larger UI change: moving several toggles under a new menu option, 'More settings.'

image

I believe the main menu is somewhat cluttered. Hiding a few items is not a bad idea.
It creates space for new features with higher prio and moves advanced options away from the average user, while maintaining sensible defaults.

Have moved the following existing items here: In-app logs, Notification settings (not rendered as a separate submenu, but as a pull-down within the same screen, as there is sufficient space), Send read receipts, and Allow Screencapture.

Additionally, I have added new: Allow saving to phone contacts, Allow deleting Vexl only contacts and Allow contact export/import.

image 
image

This menu is implemented by: src/components/MoreSettingsScreen/index.tsx

Defaults are the same for already existing options.
For new ones, all mimic the same behavior the app currently has without them (in: src/utils/preferences/index.ts), so:

Allow saving to phone contacts - ON
Allow deleting Vexl-only contacts - OFF
Allow contact export/import - OFF

**Now for the reasoning behind the new options:**

**Allow saving to phone contacts:** Since the beginning of my Vexl use, I have refused to store contacts to the device contact list.
I don't use contacts in general. If I do, there are just a few from my close inner circle and dozens Vexl contacts.
It doesn't make sense to me, and thus I really don't like the never ending 'I had to be careful not to store the contact on the phone + sometimes explain this to the counterparty'.
From my experience, I'm not alone. This make it possible to avoid entirely this flow.

The change is done by the: src/components/InsideRouter/components/SettingsScreen/components/StoreInContactList.tsx
(Changes related to this also in src/state/contacts/atom/addContactWithUiFeedbackAtom.ts + src/state/contacts/atom/addContactToPhoneWithUIFeedbackAtom.ts)

If the toggle is ON, it will allow prompting the user for storage. If OFF, the option will not be offered in the flow.

Contacts created (from QR/offer reveal) end up as Vexl-only.

This is how it looks for the remainig two:

image

**Allow deletion of Vexl only contacts:**

I'd like to show people how to use Vexl, so having the option to store contacts repeatedly would be beneficial - for training purpose. I also have some contacts I no longer want to keep. To maintain a clean contact list, I added the ability to delete contacts (the Vexl only contacts, not from the phone list) from the list/mmkv.

So far, I noticed, if contact is deleted in phone - it still remain in Vexl. I'm doing this in scope of Vexl only, this could be changed later.

It's done by src/components/SetContactsScreen/components/ContactListSelect/atom.ts (triggered from ContactItem.tsx).

According to the logs, the ensureAllImportedContactsHaveServerToClientHashActionAtom checking imported contacts for missing serverToClientHash is performed. Deletion results in lowering the connections counted in your Reach info section.

The deletion itself is slow process - don't want to make some changes around current code. I think it is enough for now, for occasional deletion. The bin icon can be hidden just by setting toggle off.

**Allow contact export/import**:

This is what bothered me most with Vexl, and many others who use it the same way. If contacts are in Vexl only storage, there is no way to transfer them to a new device (at least I don't know about). Thus losing the network with this transfer. This happened to me, and it took me a lot of time to recover same level of reach.

So I implemented import/export of .vcf files.

src/state/contacts/atom/importContactsFromVcfActionAtom.ts
src/utils/exportContactsToVcf.ts

Logic has been added to pause tasks triggered when the app goes to background/foreground. The file picker does the same.

It's done by:

src/utils/inAppLoadingTasks/useInAppLoadingTasks.ts
src/utils/useAppState.ts
src/utils/versionService/useSetupVersionServiceState.ts
src/components/NormalizeContactsWithLoadingScreen/index.tsx

The trigger reason is unclear for the user in such situation (it reloads the entire import contacts screen), so I wanted to avoid alarming user with an unnecessary reload.

Scope of the exported items is those that are checked (yellow check/to be submitted state in SetContactsScreen). Import is as Vexl only.

If no permission is granted, the first import will import contacts as Vexl only, and the permission request dialog is not visible.

However, this doesn't interfere with fetching contacts after granting permission later. A contact manually added through the search & add field is stored without issue.

Tested various import/export operations with and without contact permissions and found no problems. However, my understanding of how the backend handles this is limited.

Note that if there is overlap between phone contacts and those imported from .vcf files, and the .vcf import comes first, then when the contact is deleted, it is fetched from the phone contact list. This is the desired behavior. This situation is not expected to occur frequently because if users keep contact separate, there should be no overlap at all. But as it may still happen, I tested it just in case.

The buttons, a bit rough appearance, but clear enough with the purpose.

Basic translation is added in the packages/localization/base.json - English only as of now.

If it could be in production - can translate more in Crowdin.

There are also some logging messages around skipping background tasks, export-import, .vcf parsing, and checking for duplication, resulting in a count of the number of contacts in the .vcf versus actually imported.

////

I have no Google Maps token, so the offers detail menu displays some plaintext JSON and the map is missing. Built with just a fake placeholders in the .env. The main branch behave the same so maybe it's related.

Used the build on Android; have no iOS device to test. AFAIK, the code is for both.

Also note: the code for this PR **is AI generated**. Looking at the repo, I see a different AI was used already (Claude), so the models I'm using may be inferior. Thus, my AGENTS.md content was omitted to avoid polluting the state for the existing agent.

I haven't done any work on mobile apps for many years, last time with Kotlin, so forgive my lack of expertise here (and it's TypeScript).

I'm sure there is something missing - so let me know.

Thank you for reading this, I'm sorry it's long.


kaladivo:
Hello, 
thank you for taking your time to draft this pull request. 

You introduced some quite big UI/UX changes. We are currently in the middle of coding redesign of the app (we plan to release that before BTC Prague this year) so a lot of your changes won't be necessary. 

But we are very grateful for any feedback on the app. What would help us the most is that if you opened an GH issue for your points so we can discuss this together in more structured way :). Especially the contacts features you mentioned. 

I will close this PR for now. But please, create gh issue so we can discuss and understand your points :).

PR #2365, opened 2026-05-03
Feat/redesign/messaging

kaladivo:


kaladivo:
There seems to be some issues with reveal identity flow. Other than that, this should work!

PR #2608, opened 2026-07-14
feat: Features and fixes

SamTremko:
## Summary by CodeRabbit

* **New Features**
  * Added marketplace-ready notifications when offers become available, with notification tap navigation to the Marketplace.
  * Added clearer aggregated progress reporting while importing contacts and uploading offers.
  * Improved contact importing with smoother processing and reduced duplicate work.
  * Added new UI component demonstrations for keyboard handling, sliders, radio groups, dialogs, and swipeable offer cards.

* **Bug Fixes**
  * Improved offer-list and swipe animations.
  * Refined notification wording, identity-reveal messages, and trade reminders.

* **Localization**
  * Expanded and corrected translations across numerous supported languages, including marketplace guidance and policy content.

PR #2623, opened 2026-07-25
fix: don't re-request contacts permission on focus/resume

kaladivo:
Found while reviewing #2608.

## Bug

`useContactListSelectLifecycle` was changed to run `importContactsFromPhoneActionAtom` from `useOnFocusAndAppState` — i.e. on **every** screen focus and **every** app foregrounding (the app-state branch fires even when the screen isn't focused).

That action chains into `loadAndNormalizeContactsFromDeviceActionAtom` → `loadContactsFromDeviceActionAtom` → `getContactsAndTryToResolveThePermissionsAlongTheWay()` → `areContactsPermissionsGranted()`, which calls `requestPermissionsAsync()` whenever permission is not granted but `canAskAgain` is true.

User-visible consequence: a user who denied the contacts permission once gets the OS contacts permission prompt re-shown on every focus/resume of the Contact preferences screen. Additionally every focus/resume triggered an unthrottled full `getContactsAsync()` address-book read.

Before #2608 the focus hook only ran the passive `checkContactsAccessPrivilegesActionAtom` (`getPermissionsAsync`, never prompts), and `importContactsFromPhoneActionAtom` had exactly one caller: the explicit "import contacts" button.

## Fix

`importContactsFromPhoneActionAtom` now takes an explicit `{requestPermissions: boolean}` param:

- Lifecycle hook (focus / app resume) passes `requestPermissions: false` — it first checks the passive `areContactsPermissionsAlreadyGranted()` and skips loading entirely when access isn't already granted. `checkContactsAccessPrivilegesActionAtom` still runs via the existing `Effect.ensuring`, so the UI keeps reflecting the access state exactly as before #2608.
- The explicit "import from phone" button passes `requestPermissions: true`, keeping its current prompting behavior.

The flag is required (no default) so any future caller has to state its intent instead of silently inheriting the prompting path. This matches how the other two automatic callers guard themselves — `loadContactsFromDeviceActionAtomInAppLoadingTask` and `refreshDeviceContactsSnapshotActionAtom` both gate on `areContactsPermissionsAlreadyGranted()`.

## Verification

`pnpm turbo typecheck`, `format` and `lint` all pass for `@vexl-next/mobile-app` (only pre-existing unrelated deprecation warnings in lint).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


SamTremko:
Superseded by #2703, which carries this fix as 7e5c33c12. Everything else on this branch was already merged via #2608 (squash 70d9831ba).

PR #2624, opened 2026-07-25
fix: don't drop concurrent storedContacts writes during normalization/load

kaladivo:
Found while reviewing #2608.

## The race

Two write actions did a read-modify-write on `storedContactsAtom` with `await`s between the read and the write, so **any concurrent write to the store in that window was silently discarded**.

**1. `normalizeStoredContactsActionAtom`** — snapshots `get(storedContactsAtom)`, then normalizes in chunks of 50 per animation frame (with ~4000 contacts that is ~80 frames), then unconditionally overwrote the store with the processed *stale* snapshot.

**2. `loadContactsFromDeviceActionAtom`** — read `get(storedContactsAtom)`, then awaited `set(setDeviceContactsSnapshotFromContactsActionAtom, ...)`, then wrote the merged result. On `main` this read+write was one synchronous block; the await in between is new in #2608.

Concrete failure: normalization kicks off after app resume. Meanwhile the user finishes a vCard import (`importVexlOnlyContactsActionAtom`), adds a contact manually (`addNewContactActionAtom`), or `resolveAllContactsAsSeenActionAtom` runs. Normalization then lands its stale snapshot and those writes are gone — the user just saw an "import successful" toast and the contacts have vanished.

## The fix

Both writes now merge against the **current** store value.

- `normalizeStoredContactsActionAtom`: builds an identity `Map` from the original `StoredContact` object to its normalized replacement, then finishes with a functional update — `set(storedContactsAtom, (prev) => Array.map(prev, (c) => normalizedByOriginal.get(c) ?? c))`. Contacts added during the window stay; contacts removed during the window stay removed. Unchanged contacts keep their object identity, so the `WeakMap` identity cache in `normalizedContactsAtom` still hits. As a side effect the write no longer reorders the list (the old `Array.partition` put already-normalized contacts first).
- `loadContactsFromDeviceActionAtom`: moves `get(storedContactsAtom)` to after the last await, making the read-modify-write synchronous again. The "skip the write entirely when nothing changed" optimization and the object-identity preservation for unchanged contacts are both untouched.

Functional updates are the established pattern for this atom (`resolveAllContactsAsSeenActionAtom`, `createUpdateContactActionAtom`, `ContactListSelect/atom.ts`), and work fine through the `focusAtom` over the MMKV-persisted `contactsStoreAtom`.

## Verification

`typecheck`, `format`, `lint` all pass for `apps/mobile`; `src/state/contacts` tests pass (5/5).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/mobile/src/state/contacts/atom/loadContactsFromDeviceActionAtom.ts
+++ b/apps/mobile/src/state/contacts/atom/loadContactsFromDeviceActionAtom.ts
@@ -16,11 +16,13 @@ const loadContactsFromDeviceActionAtom = atom(null, (get, set) => {
     const contactsFromDevice = yield* _(
       getContactsAndTryToResolveThePermissionsAlongTheWay()
     )
-    const storedContacts = get(storedContactsAtom)
-
     yield* _(
       set(setDeviceContactsSnapshotFromContactsActionAtom, contactsFromDevice)
     )
+
+    // Read the store only after the last await - anything written to it while we
+    // were suspended must not be overwritten by a stale snapshot.
+    const storedContacts = get(storedContactsAtom)
 
     const contactsFromDeviceByRawNumber = new Map<string, ContactInfo>(
       Array.map(contactsFromDevice, (c) => [c.rawNumber, c])
--- a/apps/mobile/src/state/contacts/atom/normalizeStoredContactsActionAtom.ts
+++ b/apps/mobile/src/state/contacts/atom/normalizeStoredContactsActionAtom.ts
@@ -93,9 +93,9 @@ const normalizeStoredContactsActionAtom = atom(
     }
   ): Effect.Effect<void> =>
     Effect.gen(function* (_) {
-      const [toNormalize, alreadyNormalized] = pipe(
+      const toNormalize = pipe(
         get(storedContactsAtom),
-        Array.partition((contact) => !needsNormalization(contact))
+        Array.filter(needsNormalization)
       )
 
       if (!Array.isNonEmptyArray(toNormalize)) return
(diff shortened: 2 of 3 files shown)


kaladivo:
Verification review (second pair of eyes) done. Findings:

**Author's changes — verified sound:**
- `normalizeStoredContactsActionAtom`: `sequenceTasksWithAnimationFrames` is order-preserving (sequential `chunksOf` + `T.sequenceSeqArray` + flatten), so the `Array.zip(toNormalize, normalizedContacts)` identity map pairs correctly. `Array.filter(needsNormalization)` selects the exact set the old `partition` produced. In-place replacement preserves store order (the old hoist-to-front reorder was incidental; `normalizedContactsAtom` dedupe and match-sorter ranking don't depend on it). Identity-keyed merge is safe against overwriting concurrently-updated contacts: a replaced object simply misses the map and stays unnormalized, and `needsNormalization` remains true so a later run self-heals.
- `loadContactsFromDeviceActionAtom`: read-to-write is now a single synchronous continuation (no `yield*` in between); `setDeviceContactsSnapshotFromContactsActionAtom` only writes `deviceContactsSnapshotAtom`, so reading after it is fine.
- Functional updates on `storedContactsAtom` (focusAtom over the MMKV atom) are synchronous/atomic w.r.t. jotai and already used elsewhere (`resolveAllContactsAsSeenActionAtom`, `addContactWithUiFeedbackAtom`). Own MMKV persist writes don't re-decode the blob, so object identity of untouched contacts survives.

**Additional fix pushed (f2fc60e):** `updateStoredContactsAfterImportActionAtom` in `submitContactsActionAtom.ts` had the same race shape — snapshot of `storedContactsAtom`, then per-chunk animation-frame awaits (when the progress dialog is shown), then a blind overwrite. Since `updateStoredContactImportState` is a pure, cheap per-contact mapping that only depends on the fixed import results, the chunked snapshot processing is replaced with a single synchronous functional update over the current store value. Unlike an identity-keyed merge, this also applies the imported/serverHashToClient state to contacts whose objects were replaced concurrently (e.g. resolve-as-seen) and correctly handles contacts added/removed mid-import.

Also checked `ensureAndGetAllImportedContactsHaveServerToClientHashActionAtom` — it already uses the data-last `Array.map` functional-update form, so it's race-safe.

**Verification:** typecheck, format, lint (0 errors; 3 pre-existing deprecation warnings in unrelated session files), and `pnpm test src/state/contacts src/utils/vCard` (15/15) all pass.

PR #2652, opened 2026-08-05
Replace Google Maps with MapLibre + OpenFreeMap

### LINKED ISSUE #2119: Stop relying on Google Maps
Most Vexl users are privacy-minded people likely to run GrapheneOS without the optional Play Services.

The map functionality in the app should rely on the OpenStreetMap instead of Google Maps.

SamTremko:
## Why

`react-native-maps` on the Google provider ships every map screen straight to Google: the device talks to Google's tile/geocoding servers directly, using per-platform API keys we had to provision and ship in the client. That means Google sees the IP address and viewed map area of every user who opens a map screen — a direct line from device traffic to real-world location, which runs against the whole point of not being able to link user activity to real identities.

MapLibre + self-hosted style URLs removes that dependency entirely. Tile requests go to OpenFreeMap (a community-run, keyless tile provider) instead of Google, and the client no longer embeds any vendor API key.

Closes #2119.

## What changed

**Mobile** (`apps/mobile`)
- Replaced `react-native-maps` (Google provider) and `react-native-map-clustering` with `@maplibre/maplibre-react-native`. A new `VexlMap` wrapper applies the theme-aware style URL and background across all four map screens (marketplace multi-point map, single-location display, location picker, location+radius picker).
- Marketplace clustering is reimplemented with MapLibre's native GeoJSON clustering: a clustered source and a separate focused-offer source, themed pin images (light/dark, focused/unfocused), and circle/symbol layers for cluster halos, cores, and counts — replacing `react-native-map-clustering`.
- Camera control is abstracted behind a small `MapCameraControls` interface (`fitBounds`) instead of passing `MapView` refs around; region/bounds conversion between the app's `Region` shape and MapLibre's `LngLatBounds` lives in `apps/mobile/src/components/Map/utils/mapLibreRegion.ts`. Latitude is clamped to ±85° there (the web-mercator limit) so a zoomed-out region can't hand native an invalid coordinate, and bounding-box math over large coordinate arrays uses a reduce instead of spreading into `Math.min`/`Math.max`, which would risk a call-stack overflow for marketplaces with many offers.
- The client-side custom Google Maps JSON theme (~150 lines of styler rules in `mapStyle.ts`) is replaced by backend-served MapLibre style URLs (light/dark), fetched from content-service and cached in an mmkv-persisted atom with bundled fallback URLs — so the tile provider can be swapped without an app release, and the app still has a working style if the content service is briefly unreachable.
- Tapping empty map area on the marketplace map now deselects a focused offer: pin/cluster press handlers stop event propagation so only genuine "empty map" presses reach the map-level handler.
- The automatic camera recenter that used to fire when the bottom sheet was expanded/collapsed has been removed — the camera now stays completely still when the offer list is shown or hidden. Only explicit actions (initial load, search/filter changes, cluster or offer focus) move the camera.
- Removed the Google-Maps-iOS-Utils Podfile patch expo-plugin (`react-native-maps-plugin.js`) — MapLibre doesn't need it.

**Content service** (`apps/content-service`)
- New `GET /content/map-style-urls` endpoint serving the light/dark MapLibre style URLs, backed by `LIGHT_MAP_STYLE_URL` / `DARK_MAP_STYLE_URL` env config (defaulting to OpenFreeMap's positron/dark styles).

**Privacy**
- `TransformRequestManager` registers a single static, app-agnostic `User-Agent: MapLibre` header applied to every map/tile/style/glyph/sprite request, registered at app startup (`index.js`) before any map can mount. MapLibre's default User-Agent otherwise embeds the app's bundle name/identifier, which would let the tile provider single out Vexl's traffic from every other MapLibre client hitting the same shared tile host.

**Config / CI**
- Removed `IOS_MAP_API_KEY` / `ANDROID_MAP_API_KEY` from `app.config.ts`, the README env setup instructions, and the CI workflow comments that referenced them — maps need no API key now.
- Removed `react-native-maps` and `react-native-map-clustering` from `apps/mobile/package.json`; added `@maplibre/maplibre-react-native` and `@types/geojson`.

## Compatibility

This is a full swap rather than a staged rollout — there's no old/new client coexistence concern. The only environment-level change is that `IOS_MAP_API_KEY` / `ANDROID_MAP_API_KEY` are no longer required (or read) by app config evaluation, so they can be dropped from EAS env vars.

## Testing

- `pnpm turbo:typecheck`, `pnpm turbo:format`, and `pnpm turbo:lint` clean across all workspaces.
- Manual testing on iOS simulator across the marketplace map screen: initial load and fit-to-pins, cluster taps, offer focus/defocus (via pin, via tap-outside), bottom sheet expand/collapse, and light/dark theme switching.

## Summary by CodeRabbit

* **New Features**
  * Migrated mobile maps to MapLibre with improved clustering, markers, camera controls, and map interactions.
  * Added light and dark map styles, loaded dynamically from the content service.
  * Added OpenFreeMap defaults for reliable map display.
* **Documentation**
  * Updated setup guidance to clarify that map API keys are no longer required.
* **Chores**
  * Removed legacy Google Maps configuration and clustering components.
  * Improved map initialization and request identification.


kaladivo:
Will be merged as a part of #2665

PR #2656, opened 2026-08-06
Add messageAgeSeconds to MESSAGE_FETCHED_AND_REMOVED (and fix its count)

kaladivo:
## Summary

- `MESSAGE_FETCHED_AND_REMOVED` now carries a `messageAgeSeconds` attribute: the average age of the removed messages, in seconds since the server accepted them (`received_by_server_at`, which defaults to `clock_timestamp()` on insert). Falls back to `unknown` for legacy rows that predate the column.
- Fixes a pre-existing bug found while implementing this: the delete query's `WITH deleted AS (DELETE ...) SELECT count(*)` never referenced the CTE, so `count(*)` had no FROM clause and **always returned 1** — the metric's value never reflected the real number of removed messages. The DELETE now uses `RETURNING` and both the count and the average age are computed from the actual deleted rows.
- `docs/backend_stats.md` updated.

## Validation

- `pnpm turbo:typecheck`, `pnpm turbo:format`, `pnpm turbo:lint`
- chat-service test suite (71 tests) passing, including new assertions that the metric reports the real deleted count and a numeric `messageAgeSeconds`

## Summary by CodeRabbit

* **Enhancements**
  * Improved reporting when pulled messages are deleted, including the number removed and their average age.
  * Records message age when available, or marks it as unknown for older records.

* **Documentation**
  * Updated metric documentation to describe the removal count and message-age information.


kaladivo:
Folded into #2655.

PR #2665, opened 2026-08-11
feat #2664: Geocoding, reverse geocoding, and map styles served from backend services

### LINKED ISSUE #2664: Geocoding and reverse geocoding served from location service DB

SamTremko:
## Why

Two changes remove third-party dependencies from the map/location stack that PR #2652 (MapLibre + OpenFreeMap) didn't cover, plus a round of UX hardening for when geocoding/location search fails:

- **Geocoding/suggest** in `location-service` called the Google Places API server-side with `GOOGLE_PLACES_API_KEY` — every location search a user typed (and the coordinates behind a reverse-geocode) was sent to Google. New `/api/v2/geocode` and `/api/v2/suggest` endpoints serve this from a self-hosted Postgres places DB, seeded from OSM + Natural Earth data, so that traffic never leaves our infrastructure. **The old `/geocode` and `/suggest` endpoints stay Google-backed** for already-released app clients that still call them — see Compatibility below.
- **Map style documents** were fetched by the mobile client directly from the style provider (content-service only handed back the light/dark URLs). content-service now fetches, validates and caches the style JSON itself and returns it inline via `GET /content/map-styles`, so the client only ever talks to our backend for it.
- Geocoding/location search can fail (no network, service down, no match). The mobile app now surfaces that instead of failing silently: loading/error states, retry with backoff for transient failures, and a manual-location fallback (type an address, or for filters, just use the map area) when it can't resolve automatically.

Closes #2664.

## What changed

**Geocoding DB** (`packages/geocoding-db`, new package)
- Standalone Postgres-backed places DB, isolated from the shared `DB_URL` vexl database (own `GEOCODING_DB_*` env / `geocoding-postgres` instance).
- `GeocodingDbService` with `createQueryNearestPlace` (reverse geocoding) and `createQuerySuggestPlaces` (two-phase: important-places-first, then typo-tolerant trigram fallback for suggestions).
- `common.ts` holds the shared vocabulary (settlement/POI type weights, importance scoring, viewport radius, diacritic-folding name normalization) used identically by the query layer and by ingest, so search-time and ingest-time data can't drift.
- Ingest pipeline (`scripts/ingest.ts`, `scripts/ingestParsing.ts`) for OSM/Natural Earth → Postgres, `scripts/refresh.sh` for the manual (operator-run, no server-side cron) dataset refresh, `scripts/seedDev.ts`/`devSeedData.ts` for a small local dataset.
- Migration (`src/migrations/0001_initial.ts`), Jest specs for parsing/normalization/refresh, and a `README.md` documenting setup, ingest, and refresh.

**Location service** (`apps/location-service`)
- New `/api/v2/suggest` and `/api/v2/geocode` serving `GeocodingDbService`; the `rest-api` client wrapper targets only the v2 endpoints, so updated mobile clients never call v1.
- v1 `/suggest` and `/geocode` are unchanged — still backed by `GoogleMapsService`/`GOOGLE_PLACES_API_KEY`, kept intentionally so already-released clients keep behaving exactly as before until v1 traffic can be retired.
- Mobile's location search/geocoding now sends the current app language through to v2, so suggestions/labels come back localized.

**Content service** (`apps/content-service`)
- New `GET /content/map-styles` endpoint (`getMapStylesHandler`/`MapStylesService`) replacing `GET /content/map-style-urls`: fetches the light/dark MapLibre style documents from the configured URLs, validates they look like a style document, and returns them inline as `MapStylesResponse` (`{light, dark}` JSON strings) instead of bare URLs.
- Response is cached in Redis (`CacheService`) alongside events/blogs, same 1h TTL, cleared by the existing `clearCache` path. Upstream fetch failures are logged with a sanitized cause (no raw URLs/headers).
- `rest-api` content-service contracts/specification updated: `MapStyleUrlsResponse` → `MapStylesResponse`, new branded `MapStyleJson` string type.

**Mobile** (`apps/mobile`)
- `mapStyleUrlsAtoms.ts` → `mapStylesAtoms.ts`: the mmkv-persisted atom now stores the fetched style JSON (with bundled fallback documents) instead of a URL, consumed directly by `VexlMap`. Loading task renamed to `loadMapStylesInAppLoadingTask.ts` to match.
- New `localizeMapStyleLabels.ts` localizes the light/dark style names read out of the fetched JSON, since the label text now comes from the backend response rather than being hardcoded client-side.
- **Manual location fallback and retry** (`LocationPicker`, `MapLocationWithRadiusSelect`, `LocationSearch`): when geocoding/search fails, users can type an address manually instead of being stuck; transient failures get retried with backoff (`transientRequestRetryPolicy.ts`).
- **Filter location works with broken geolocation too**: `LocationRadiusPicker` gained an `offer`/`filter` variant — in the filter flow, manual entry can fall back to just the picked map area (no address text required) instead of forcing address input.
- **Failures are surfaced, not swallowed**: loading/error states on location search and reverse geocoding across `LocationMapSelect`, `LocationSearch`, `MapLocationSelect`, `MapLocationWithRadiusSelect`; a failed re-geocode of an already-picked location no longer clears the selection or blocks saving. `reportLocationServiceError.ts` reports only the error tag and status code (never the request, which embeds the searched coordinates/phrase).
- New reusable `LengthCounter` UI component (character counter for text fields with a max length), used by the manual-address input; `TextField`/`LabeledTextArea` updated to support it; new ui-book screens.

## Compatibility

- **location-service**: v1 `/geocode`/`/suggest` are unchanged and still require `GOOGLE_PLACES_API_KEY` — no breaking change for released clients. v2 (`/api/v2/geocode`, `/api/v2/suggest`) is new and requires an ingested geocoding DB (`GEOCODING_DB_*` env). v1 will be deleted once v2 is proven reliable on new clients.
- **content-service**: old `/content/map-style-urls` endpoint is gone in favor of `/content/map-styles` — this is a breaking change for that one endpoint, but it's only ever called by the mobile client shipped alongside this change.

## Testing

- `pnpm turbo:typecheck`, `pnpm turbo:format`, and `pnpm turbo:lint` clean across all workspaces.
- Unit tests: `geocoding-db` ingest/parsing/normalization/refresh specs; `location-service` v1 (mocked Google) and v2 geocode/suggest specs against a seeded test DB; `content-service` `mapStyles` route test.
- Manual testing: location search/geocoding (suggestions, reverse geocode, manual fallback, filter-with-broken-geolocation) and map style light/dark loading in the mobile app.

## Summary by CodeRabbit

* **New Features**
  * Maps now use MapLibre with OpenFreeMap tiles and localized light/dark styles—no map API key required.
  * Added improved location suggestions and reverse geocoding, including multilingual search, typo tolerance, and better place results.
  * Added manual address entry and map-based fallback when location services are unavailable.
  * Added character counters to text fields with maximum-length support.
* **Bug Fixes**
  * Improved map clustering, camera fitting, retry handling, and location-search error recovery.
* **Documentation**
  * Updated setup, location-service, local development, and geocoding database documentation.


jarvisonsmart:
> 🤖 [Fable] on Dave's behalf.

## Reverse geocoding now resolves by boundary containment — worldwide (`9038aa1c7`)

**The problem.** `/geocode` picked the *nearest settlement label node* and used it as the label. Point distance is not containment: a pin in Holešovice (Praha 7) returned "Josefov, Praha" because the Josefov node is 755 m away and the Holešovice node farther. Any pin near a district edge got the neighbour's name; in countries where OSM maps neighbourhoods as nodes only, the "city" context could also be a random nearby city node.

**What changed**

*Data (`tools/location-db-updater`)*
- `refresh.sh` now extracts boundary polygons from the raw extracts: `boundary=administrative` relations at admin levels 6–11 (plus any level when the relation carries a settlement `place=*` tag — Berlin, Wien, Praha), `place=*` settlement polygons (ways + relations), and `boundary=cadastral` relations (kept for CZ only — that's what gives "Holešovice").
- Boundaries are stored with `name`, translations, `country_code` (Natural Earth centroid lookup), `boundary_type`, `admin_level`, `place_tag`. Geometry is run through `ST_MakeValid` → `ST_SimplifyPreserveTopology` at 0.0001° (~10 m) → `ST_Subdivide` (≤ 256 vertices/part) into a GiST-indexed parts table. New migration `0002_placeBoundaries`; the dev dump (`tooling/dev/geocoding-postgres-init`) is rebuilt with the new schema (21 MB).

*Query (`packages/geocoding-db`)*
- Each boundary gets a **role resolved at query time** from `(country_code, admin_level, place_tag)` — `common.ts#boundaryRole`: place tag wins (city/town/municipality → city, suburb/village/… → sub-city), CZ cadastral → sub-city, level 8 → city, 9–11 → sub-city, ≤ 7 ignored, plus per-country overrides (`COUNTRY_ADMIN_LEVEL_ROLES`: DK/NO/SE kommune at 7, PT concelho/freguesia at 7/8, AT Katastralgemeinden at 10 ignored). Fixing a country's semantics is a one-line change with no re-ingest.
- Resolution: candidates = boundaries within ~20 m of the pin (covers slivers between independently simplified neighbours), covering first, smallest area next. The most specific sub-city boundary is the label; the covering city/town boundary is the "…, City" context, otherwise the best city/town node within 30 km (importance − 0.005/km, same country first). A city boundary with no sub-city polygon inside falls back to the nearest sub-city *node inside that city* within 1.5 km (US-style neighbourhood nodes), then the city itself; with no boundary at all the previous nearest-settlement lookup still applies.
- One intentional UX change: a `municipality`/untagged level-8 polygon is no longer a self-sufficient label (a Czech obec can be one village) — only cities/towns are, so "Černilov, Hradec Králové" stays as before. Response shape is unchanged.

**Measured with the real pipeline against local PostGIS**
| dataset | boundaries | geometry (incl. GiST) | notes |
|---|---|---|---|
| Czechia | 21.7k | 61 MB | was 174 MB exact in the WIP |
| Europe | 485k | ~1 GB | |
| World (all 8 Geofabrik continents) | 1.07 M | 2.2 GB + 122 MB metadata | whole DB 10 GB; 0 invalid parts; lookups 1–20 ms |

Pins checked: 22 in CZ (Josefov case fixed, "Holešovice, Praha"), 40 across SK/DE/PL/AT/GB/FR/IT/NL/ES/HU/PT/NO/SE/DK/CH/BE, 51 worldwide (NYC, SF, Toronto, CDMX, São Paulo, Buenos Aires, Tokyo, Seoul, Mumbai, Bangkok, Sydney, Cape Town, Istanbul, Dubai, Moscow, …) — all read as sensible rough locations; a few OSM naming quirks remain ("Manhattan Community Board 7, New York", "RW 05, Central Jakarta").

**Tests**: role mapping + candidate resolution units, containment integration (specificity, slivers, node-in-city fallback, municipality context, country overrides, importance/distance/same-country context), ingest parsing, end-to-end pipeline (simplification, subdivision, `ST_MakeValid` repair, reverse geocoding through ingested data), `refresh.sh` filters. `turbo:typecheck/format/lint` green. READMEs of `geocoding-db`, `location-db-updater` and `location-service` describe the data tier, the role table and the tolerances.

Rollout note: the first prod refresh after merge ingests the worldwide boundaries (~2.3 GB extra); the existing atomic swap + 30 % sanity gate apply.

jarvisonsmart:
> 🤖 [Fable] on Dave's behalf.

Follow-up (`c06f42371`, force-pushed over `9038aa1c7`): compared v1 (Google) vs v2 (geocoding DB) on 20 Czech pins — all 20 resolve to the same area (v2 is usually finer where Google returns "7, Kladno"-style numbered units: "Kročehlavy, Kladno", "Mariánské Hory, Ostrava", "Bedřichov u Jihlavy, Jihlava").

One defect found and fixed: a pin in Český Těšín read "Kontešinec, Český Těšín - **PL**". The boundary's country came from a Natural Earth lookup of its bbox centre at ingest, and for a district hugging the Olza river even a true interior point lands on the Polish side of NE's ~km-accurate border. Two-layer fix:
- ingest looks the country up from an interior point of the largest polygon (scanline midpoint, `ingestParsing.ts#interiorPoint`) instead of the bbox centre;
- the query reports the country of the **largest covering boundary** (nested boundaries always share a country; the okres/obec around a border district is never misattributed) — `resolveBoundaryCandidates` → `countryCode`.

Tests for both; CZ dev dump rebuilt (136,695 places / 21,692 boundaries) and restore-verified; typecheck/format/lint green.

kaladivo (approved):

PR #2703, opened 2026-08-31
fix #2635, #2636, #2637, #2638, #2623, #2621: Trade checklist staged-update scoping, marketplace scroll reset, exit navigation dedupe, progress dialog polish, contacts permission re-prompt, contacts backup cleanup timing

### LINKED ISSUE #2621: Contacts backup export: .vcf deleted immediately after shareAsync resolves — lazy share targets may read an empty file
Found during review of #2608.

`exportVexlOnlyContactsActionAtom` (`apps/mobile/src/state/contacts/atom/exportVexlOnlyContactsActionAtom.ts`) deletes the backup `.vcf` in a `finally` the moment `shareAsync` resolves. On Android the share hands out a `content://` URI, and common targets (Gmail attach, Drive upload) read it lazily after the chooser returns — the delete can yield an empty or failed attachment.

The privacy intent (don't leave a plaintext contact dump in cache) is right; just defer the delete — e.g. to screen unmount or next app resume — and verify against real share targets (Gmail, Drive, Files) on Android.

### LINKED ISSUE #2635: Trade checklist: staged updates are global across chats and survive hardware-back — stale identity/contact reveal can be sent to the wrong chat
`updatesToBeSentAtom` (`TradeChecklistFlow/atoms/updatesToBeSentAtom.ts`) is a single module-global atom, not scoped per chat, and `setParentChatActionAtom` (`state/tradeChecklist/atoms/fromChatAtoms.ts`) does not reset it when the checklist opens for a different chat. It is only cleared on a successful send or via the AgreeOnTradeDetails X-button confirm dialog.

Escape hatch: Android hardware back (or iOS swipe-back) on `AgreeOnTradeDetailsScreen` pops the whole `TradeChecklistFlow` without any `beforeRemove` guard, so staged updates silently persist.

Failure scenario: user stages an identity reveal (and optionally contact reveal = **full phone number**) in chat A's checklist, exits via hardware back, later opens chat B's checklist. `areThereUpdatesToBeSentAtom` is true, the Send button is enabled, and pressing it transmits chat A's staged identity/contact data to chat B's counterparty. Given the privacy-first principle, wrong-recipient delivery of identity data is the worst-case class of bug.

Suggested fixes (complementary):
1. Add a `beforeRemove` guard on AgreeOnTradeDetails that routes through the existing `askAreYouSureAndClearUpdatesToBeSentActionAtom`.
2. Structurally: scope staged updates per chat (or clear in `setParentChatActionAtom` when the target chat changes) so stale updates can never cross chats regardless of exit path.

Note: branch `fix-2616-board_refinement` adds a related safeguard for the note-chat reveal path (clearing staged updates on failed submit), but the checklist-origin path described here is pre-existing and unaffected.

### LINKED ISSUE #2636: Marketplace: programmatic tab-switch scroll-to-top runs post-paint, causing a visible 1–2 frame jump
On branch `fix-2616-board_refinement`, `MarketplaceScreenContent.tsx` replaced the pre-paint `useLayoutEffect` scroll reset with the new `useScrollToTopOnFocus` hook (focus effect + one rAF). Because both tab scenes are retained (`RetainedScene`), navigating to the marketplace with a `tabSwitchRequestId` (e.g. after creating an offer → myOffers tab) now shows the newly active scene at its stale scroll offset for a frame or two before visibly jumping to the top. Previously the reset happened before paint with no flash.

If the post-focus timing was needed for correctness (e.g. FlashList not measured yet on first focus), consider scrolling in a `useLayoutEffect` keyed on the requestId when the screen is already focused, and only falling back to the focus-effect path otherwise.

Related latent trap in the same file: `activeTabScrollRequestId` deliberately drops `scrollToTopRequestId` for the `allOffers` tab, so a future `navigate(..., {initialTab: 'allOffers', tabSwitchRequestId})` would silently not scroll. All current callers pass `'myOffers'`, so it is not user-visible today — but the asymmetry deserves a comment or removal.

### LINKED ISSUE #2637: Trade checklist: six copies of the submit-then-exit navigation block across sub-flows
The "was this flow opened from AgreeOnTradeDetails → either park updates and popTo the checklist, or submit immediately and popTo ChatDetail" block is duplicated across the checklist sub-flows:

- `CalculateAmountFlow/components/ConfirmAmountScreen/index.tsx` (~81–89)
- `NetworkFlow/components/NetworkScreen/index.tsx` (~42–50)
- `DateAndTimeFlow/components/PickTimeFromSuggestions/index.tsx` (~145)
- `MeetingLocationFlow/.../LocationMapSelect.tsx` (~85)
- `MeetingLocationFlow/.../LocationMapPreview.tsx` (~47)
- `RevealIdentityFlow/useRevealIdentityFlowNavigation.ts` (branch `fix-2616-board_refinement`)

Each copy also manually wraps the submit in `loadingOverlayDisplayedAtom` even though `submitTradeChecklistUpdatesActionAtom` already wraps the send in `withLoadingOverlayAtom` — the outer overlay is redundant (and adds a one-tick flash on the "no updates"/"cannot send" fast paths).

A shared `useTradeChecklistExitNavigation()` hook (combining `useWasOpenFromAgreeOnTradeDetailsScreen` + submit + popTo, without the redundant overlay) would replace all six and would also resolve the deep-entry back/close bug tracked in #2634. None of the copies handle submit failure consistently either (staged updates linger — see #2635), so centralizing is the right place to fix that once.

### LINKED ISSUE #2638: ProgressDialog: indicateProgress {type: 'success'} actually means "hide the indicator" — confusing next to {type: 'done'}
In `packages/ui/src/components/ProgressDialog.tsx`, the `ProgressIndication` union has both `{type: 'done'}` and `{type: 'success'}` (added on branch `fix-2616-board_refinement`). Their names suggest the same thing, but they behave differently:

- `'done'` → renders a full progress bar
- `'success'` → hides the progress indicator (and `belowProgressLeft/Right`) entirely

A name like `{type: 'none'}` / `'hidden'` — or making `indicateProgress` optional — would state the intent. Currently the only consumer of `'success'` is CreateNoteScreen's posted-toast step, so the rename is cheap now and gets more expensive with every new consumer.

SamTremko:
Closes #2635, closes #2636, closes #2637, closes #2638, closes #2621. Also lands the fix from #2623 (supersedes it).

## #2635 — Trade checklist staged updates are global across chats and survive hardware-back

Staged checklist updates (including identity reveal and full phone number) lived in one module-global atom, were only cleared on a successful send or the X-button dialog, and survived hardware-back — so a stale reveal could be sent to a different chat's counterparty.

- **Structural scoping**: the base `updatesToBeSentAtom` (+ `areThere…`/`clear…`) moved to a dependency-light `state/tradeChecklist/atoms/updatesToBeSentAtom.ts`, and `setParentChatActionAtom` now clears staged updates whenever the parent chat changes. No caller can forget the clear anymore (replaces the per-call-site clearing).
- **Exit guard**: `TradeChecklistFlow` uses the shared `usePreventDiscardChangesWithConfirmation` hook routed through the existing discard dialog, covering hardware back / swipe / any removal. The X-button simplifies to a plain `popTo`; the send path auto-passes because nothing is staged after a successful submit.
- **Refocus re-sync**: `TradeChecklistFlow` and `MessagesScreen` sync the parent chat on focus (not just mount), so a notification opening another chat on top can't leave the visible checklist writing to the wrong chat. Re-setting the same chat is a no-op, so same-chat drafts are preserved.
- New regression test `fromChatAtoms.test.ts` pins the invariant (switching chat clears staged updates; same chat keeps them). The obsolete jest mock in `revealIdentityAtoms.test.ts` is removed since the base atom module is now jest-safe.

## #2636 — Marketplace tab-switch scroll-to-top runs post-paint

At HEAD the pre-paint `useLayoutEffect` on `activeTab` still existed, so the residual defect was the same-tab case (already on myOffers → create/pause offer → back to myOffers): only `useScrollToTopOnFocus` (focus effect + rAF) reset the scroll, one frame after paint. The `activeTabScrollRequestId` asymmetry mentioned in the issue never landed in any commit.

- `scrollToTopRequestId` is folded into the existing layout effect, so a request for the already active tab resets before paint.
- The params → `activeTab` sync in `MarketplaceScreen` is a layout effect, so a programmatic tab switch and its scroll reset land in one paint (child layout effects flush before the parent's).
- `BoardScreen` (the hook's only other consumer) gets the same shape for filter-switch requests; `useScrollToTopOnFocus` is removed. Net −41 lines.
- Focus-gating fallback verified unnecessary: Marketplace is the always-mounted initial tab, nothing in the app freezes/unmounts on blur, and Board's lazy first mount is already at offset 0.

## #2637 — Six copies of the submit-then-exit navigation block

Extracted into the shared `useTradeChecklistExitNavigation` hook used by all sub-flows.

## #2638 — ProgressDialog `indicateProgress {type: 'success'}` semantics

Progress indication polishing for notes so the success/done states read consistently.

## #2623 — Contacts permission re-prompted on every focus/resume

`useContactListSelectLifecycle` runs `importContactsFromPhoneActionAtom` on every screen focus and app foregrounding, and that chained into `requestPermissionsAsync()` whenever permission was denied but `canAskAgain` — so a user who denied once got the OS contacts prompt again on every focus/resume of Contact preferences.

- `importContactsFromPhoneActionAtom` takes a required `{requestPermissions: boolean}`. With `false` it checks the passive `areContactsPermissionsAlreadyGranted()` (`getPermissionsAsync` only) and skips loading when access isn't already granted; `checkContactsAccessPrivilegesActionAtom` still runs via the existing `Effect.ensuring`, so the access-state UI is unchanged.
- Lifecycle hook passes `false`; the explicit "import from phone" button passes `true` and keeps prompting.
- Same shape as the other automatic callers (`loadContactsFromDeviceActionAtomInAppLoadingTask`, `refreshDeviceContactsSnapshotActionAtom`), which already gate on the passive check.
- Not in scope: granted users still get an unthrottled address-book read on every focus/resume — that import-on-focus was added deliberately by #2608, so throttling it is a separate decision.

## #2621 — Contacts backup .vcf deleted before Android share targets finish reading it

`exportVexlOnlyContactsActionAtom` wrote the vCard backup file, awaited `shareAsync`, then deleted the file in a `finally`. On Android, lazy share targets (Gmail attach, Drive upload) read the `content://` URI *after* `shareAsync` resolves, so the delete could race ahead and hand them an empty file.

- Deletion is pulled out into `cleanupVexlOnlyContactsBackupFileActionAtom` and moved off the share call path entirely — it now runs on next screen focus, app resume, and screen blur/unmount (`VexlOnlyContactsScreen`), so the file outlives `shareAsync` but doesn't linger indefinitely.
- `deleteFileIfExists` is a small best-effort helper shared between the pre-write cleanup and the deferred cleanup.

## Verification

`pnpm turbo:typecheck`, `pnpm turbo:format`, `pnpm turbo:lint` all pass; the full mobile jest suite passes (38 suites, 296 tests). The full `turbo:test` run only fails in `notification-service` with `PgClient: Failed to connect` (no local Postgres) — unrelated to these mobile-only changes. #2636 frame timing is not device-verified; #2623 is worth a device check (deny permission → focus/resume twice → no prompt; grant in Settings → contacts still refresh on focus); #2621 is worth a device check too (share the vCard to Gmail/Drive on Android → attachment isn't empty).

## Summary by CodeRabbit

- **New Features**
  - Improved trade checklist exit flow, including safer handling of unsaved changes and consistent return navigation.
  - Contact imports now request phone permissions only when initiated by the user.
  - Contact backup exports remain available while sharing completes in the background.

- **Bug Fixes**
  - Trade checklist updates are cleared when switching to another chat.
  - Chat details and marketplace/board screens now restore focus, filters, and scroll positions more reliably.
  - Note confirmation dialogs no longer display an unnecessary success indicator.


kaladivo (approved):

SamTremko:
@Greptile review

SamTremko:
@greptile review

PR #2707, opened 2026-09-02
fix #2620: Chunk vCard import work and dedupe against stored contacts

### LINKED ISSUE #2620: vCard import: synchronous parsing can freeze the JS thread (ANR); dedupe misses in-flight contacts
Found during review of #2608.

1. **Synchronous parse/normalize freeze.** `importVexlOnlyContactsActionAtom` (`apps/mobile/src/components/ContactPreferencesFlow/components/ContactListSelect/atom.ts`) parses the whole `.vcf` and runs `toE164PhoneNumberWithDefaultCountryCode` over every TEL line in a plain `for` loop with no yield points. `MAX_CONTACTS_PER_VCF_IMPORT = 5000` is applied *after* normalization, so the only real bound is `MAX_VCF_FILE_SIZE_BYTES = 2 MB` ≈ ~40,000 minimal vCard entries. Then up to 5,000 contacts go through `hashPhoneNumberE` (sync HMAC) + `Schema.decodeSync` in one `Effect.forEach`. Picking a large `.vcf` freezes the JS thread for seconds before the confirm dialog appears — ANR territory on Android. `normalizeStoredContactsActionAtom` already chunks identical work at 50 items per animation frame; the import should reuse that pattern.

2. **Dedupe uses normalized contacts only.** The import dedupes against `normalizedContactsAtom`, which excludes contacts with `computedValues: None` or `invalidNumber: 'invalid'`. Importing a `.vcf` while first normalization is still in flight persists duplicate `StoredContact` rows (hidden by the UI but they survive `eraseImportedContacts`, and after submit `importedContactsHashesAtom` yields the same hash twice). Dedupe against `storedContactsAtom` (rawNumber + normalizedNumber when present) instead.

3. (Nice to have) The parser ignores `ENCODING=QUOTED-PRINTABLE`, so vCard 2.1 exports (Android legacy composer) with non-ASCII names import garbled. Round-tripping Vexl's own 3.0 export is unaffected.

SamTremko:
Closes #2620.

- **No more JS-thread freeze on large `.vcf` files:** `importVexlOnlyContactsActionAtom` flattens parsed contacts into per-TEL entries and normalizes them in chunks of 50 with an animation-frame yield per chunk (the same pattern `normalizeStoredContactsActionAtom` uses — the chunk size is now a shared constant in `contactImportUtils.ts`). The `MAX_CONTACTS_PER_VCF_IMPORT` cap is checked before normalizing, and the loop stops entirely once it's reached, counting the remainder as skipped. The post-confirm hash+decode step is chunked the same way.
- **Dedupe against `storedContactsAtom`:** the existing-numbers set is built from every stored contact's `rawNumber` plus `normalizedNumber` when present, so importing while first normalization is still in flight can no longer persist duplicate rows.
- **Quoted-printable support (the nice-to-have):** `parseVcardString` recognizes `ENCODING=QUOTED-PRINTABLE` (incl. the bare vCard 2.1 form), joins QP soft line breaks, and decodes values as UTF-8 with a raw-value fallback on malformed input, so legacy Android vCard 2.1 exports with non-ASCII names import correctly. Covered by new unit tests, including a regression guard for literal `=` in non-QP cards.

## Test plan

Import the same `.vcf` twice: the first run should import, the second should report "nothing new". A vCard 2.1 file with QP-encoded diacritics in names should import them un-garbled.

## Summary by CodeRabbit

- **New Features**
  - Improved contact imports by excluding duplicates, including contacts stored concurrently during the import flow.
  - Added support for quoted-printable encoded values in vCard files, including multiline and structured fields.
  - Added a success confirmation after exporting contacts, with guidance on restoring the `.vcf` backup.

- **Bug Fixes**
  - Improved resilience when reading malformed encoded vCard data.
  - Prevented malformed vCard lines from interfering with subsequent contact fields.

- **Chores**
  - Updated project ignore rules for generated Android build files.
  - Added translations for the contact export confirmation.

### DIFF
--- a/apps/mobile/src/components/ContactPreferencesFlow/components/ContactListSelect/atom.ts
+++ b/apps/mobile/src/components/ContactPreferencesFlow/components/ContactListSelect/atom.ts
@@ -12,7 +12,10 @@ import {atomFamily, splitAtom} from 'jotai/utils'
 import {matchSorter, rankings} from 'match-sorter'
 import {Linking} from 'react-native'
 import {addContactToPhoneActionAtom} from '../../../../state/contacts/atom/addContactToPhoneWithUIFeedbackAtom'
-import {CONTACT_IMPORT_PROGRESS_DIALOG_MIN_CONTACTS} from '../../../../state/contacts/atom/contactImportUtils'
+import {
+  CONTACT_IMPORT_PROGRESS_DIALOG_MIN_CONTACTS,
+  CONTACT_NORMALIZATION_CHUNK_SIZE,
+} from '../../../../state/contacts/atom/contactImportUtils'
 import {
   normalizedContactsAtom,
   storedContactsAtom,
@@ -22,6 +25,7 @@ import {submitContactsActionAtom} from '../../../../state/contacts/atom/submitCo
 import {
   StoredContactWithComputedValues,
   type ContactsFilter,
+  type StoredContact,
 } from '../../../../state/contacts/domain'
 import {
   areContactsPermissionsAlreadyGranted,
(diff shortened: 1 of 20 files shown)


SamTremko:
@greptile review

SamTremko:
@greptile review

SamTremko:
@greptile review

kaladivo (approved):

Issue #2622, opened 2026-07-25
Perf follow-ups from #2608: contact filter atom subscriptions, double splitAtom, ineffective resume throttle, intro GIF cost


kaladivo:
Perf watch-list from review of #2608 — worth measuring, since the PR's goal was perf improvement:

1. **Filtered contact-atom lists subscribe to every row atom.** `createFilteredContactAtomsAtom` (`ContactListSelect/atom.ts`) calls `get(contactAtom)` for all n split atoms; at 10k contacts that's three derived atoms each with 10k dependency edges, rebuilt whenever search results change. Combined with all four FlashLists staying mounted via `RetainedScene` (opacity 0), a single contact edit invalidates and recomputes all three filtered arrays. On main, per-filter atoms had two dependencies each.

2. **Two overlapping `splitAtom`s over my-offers.** `myOffersSortedAtomsAtom` (`apps/mobile/src/state/marketplace/atoms/myOffers.ts`) survives solely for `MyOffersListHeader` (which appears to need only a count) alongside the new `visibleMyOffersAtomsAtom` — every my-offers change churns two per-offer atom caches.

3. **The new 5-minute `minTimeBetweenRunsMs` throttle on `refreshOffersAndEnsureInboxesInAppLoadingTask` is mostly bypassed:** `AllOffersScene` still runs an unthrottled `Effect.runFork(refreshOffers())` from `useAppState` on every resume (`MarketplaceScreenContent.tsx`), and that scene is now permanently mounted. Two concurrent `refreshOffersActionAtom` runs on one resume remain possible (no in-flight guard; the merge itself is safe).

4. **Marketplace intro GIF**: single 320×422 GIF per theme (132/170 frames, ~858 KB / ~971 KB) rendered at full dialog width with `objectFit="fill"` — upscaled ~3x on 3x devices (soft where the old @3x PNG was sharp) and a real decode cost on low-end Android.

PR #2723, opened 2026-09-04
perf: avoid building offer atoms just to display a count

jarvisonsmart:
The My offers header subscribed to a separate per-offer atom list solely to display its length. It now reads a scalar count from the owned offers, allowing the redundant splitAtom cache to be removed. The existing list atoms, active/paused sections, and sorting remain unchanged.

Addresses item 2 of #2622. The other performance investigations remain open.

Verification:
- Mobile typecheck; repository format and lint pass. Existing deprecation warnings remain.
- Executed the actual offer atom and sorting modules with real Jotai, optics, and Effect: foreign offers are excluded; active and paused owned offers count; sorting and pausing do not notify count subscribers; list order and offer atom identity are preserved; deleting the last owned offer gives zero.
- Android API 37 emulator: rendered the production header with real offer state and local fixtures. UI assertions and screenshots verify three owned offers across sorting/pause, one owned offer, and zero owned offers hiding the header. Temporary test entry removed afterward.

### DIFF
--- a/apps/mobile/src/components/InsideRouter/components/MarketplaceScreen/components/MyOffersListHeader.tsx
+++ b/apps/mobile/src/components/InsideRouter/components/MarketplaceScreen/components/MyOffersListHeader.tsx
@@ -3,7 +3,7 @@ import {useAtom, useAtomValue} from 'jotai'
 import React, {useCallback} from 'react'
 import {Stack, useTheme, XStack} from 'tamagui'
 import {
-  myOffersSortedAtomsAtom,
+  myOffersCountAtom,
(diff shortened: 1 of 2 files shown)


jarvisonsmart:
> 🤖 [GPT-6] on Dave's behalf.

I checked the docstring-coverage warning and am leaving the code unchanged. AGENTS.md asks us to avoid comments that duplicate the code. The count atom reads the owned-offers length, and the header consumes that count; neither needs a docstring to explain an unclear constraint. The local checks and Android emulator verification cover count changes, sorting, paused offers, and the zero state.

kaladivo (approved):

PR #2740, opened 2026-09-11
feat(mobile): #2550 refresh offer common friends when the social graph changes

### LINKED ISSUE #2550: Update common friends for already created offers
Once the offer is created and common friends are set, they are not changed in the offer once user imports more common contacts with the counterparty.

BLOCKED BY faster crypto

SamTremko:
Closes #2550

## Problem

An offer's private part for a recipient is encrypted once, when that recipient first becomes a connection. It carries the common friends and connection level at that moment and is never touched again, so recipients see stale common friends as soon as either side's contacts change.

## What it does

The offer refresh now fetches the social graph itself, diffs it against the stored one, and re-uploads private parts for the recipients whose common friends, verified friends, or connection level changed. New and removed recipients are handled as before. The server replaces the old private parts, so recipients see the current common friends after the next refresh round.

`updatePrivateParts` gains a `connectionsToRefresh` input for those recipients and returns `updateSuccess`, which is false when the background deadline skipped recipients or an upload batch failed. Recipient keys that cannot be encrypted for are reported as before but do not count as a failed round.

## Design notes

- The fetched graph is stored only once every offer reports `updateSuccess`. A round cut short by the 25 second background budget keeps the old graph, so the next round computes the same diff and re-uploads it. This avoids any bookkeeping of what reached the server.
- Callers that previously ran the connections sync before the offer refresh drop that step. Storing the fresh graph first would make the diff empty and skip the re-encryption. The app-loading task runs the import-needed check after the refresh so it still sees the fresh graph.
- The manual offer edit fetches the graph for its own upload but does not store it, leaving the other offers' diff intact for the next round.
- If the fetch fails, the refresh falls back to the stored graph with nothing to re-encrypt, matching the previous behaviour when the sync failed.

## Risks

- After a partial round, if a recipient's data changes and then reverts before the next successful round, the offers uploaded in the partial round keep the intermediate common friends until that recipient changes again. Judged acceptable given the rarity and that the list was never refreshed before.
- Background rounds with very large networks may need several attempts before a diff-heavy round completes and the graph is stored. Each attempt re-uploads the same diff.

### DIFF
--- a/apps/mobile/src/components/DebugScreen/index.tsx
+++ b/apps/mobile/src/components/DebugScreen/index.tsx
@@ -82,9 +82,7 @@ import {upsertInboxOnBeAndLocallyActionAtom} from '../../state/chat/hooks/useCre
 import {clubsToKeyHolderAtom} from '../../state/clubs/atom/clubsToKeyHolderV2Atom'
 import {clubsWithMembersAtom} from '../../state/clubs/atom/clubsWithMembersAtom'
 import {type ClubWithMembers} from '../../state/clubs/domain'
-import connectionStateAtom, {
-  syncConnectionsActionAtom,
-} from '../../state/connections/atom/connectionStateAtom'
+import connectionStateAtom from '../../state/connections/atom/connectionStateAtom'
 import offerToConnectionsAtom, {
   deleteOrphanRecordsActionAtom,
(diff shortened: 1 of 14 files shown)


SamTremko:
Superseded by #2741 after renaming the branch to feat-2550-update_offers_common_friends.

PR #2741, opened 2026-09-11
#2550 Refresh offer and note common friends with responsive, retryable sync

### LINKED ISSUE #2550: Update common friends for already created offers

SamTremko:
Closes #2550

## Problem

Existing offer and note recipients keep stale common-friend information when the social graph changes. Partial refreshes need durable retries, and removed repost recipients must lose server access. Large encryption batches can block the UI and delay other loading tasks; advancing note pagination before storing notes can also skip results after an interrupted refresh.

## What it does

- Refreshes encrypted private parts for existing V2 recipients when common friends or connection levels change; offers also refresh verified-friend information. Normal recipient additions and removals continue to support both key versions.
- Tracks pending refreshes per offer and per note, retaining failed work while clearing successful records. Notes own their graph baseline and lock independently of offers, so an offer refresh cannot consume note changes.
- Replaces note private parts transactionally within the existing upload endpoints, keeping direct shares and individual reposts separate and preventing duplicate rows on retries.
- Adds scoped removal of recipients from a repost, restores an empty repost when it gains recipients again, and removes local tracking for notes or reposts that are no longer available on the server.
- Reports note encryption failures, upload failures, and deadline skips as incomplete updates. Repost payloads continue to omit common friends and specific connection levels.
- Yields between offer and note encryption time slices, reuses recipient planning, and avoids unnecessary connection-record writes. Connection sync runs after other selected loading tasks finish or are skipped.
- Shows a localized offer-encryption status after one second, keeps it active through uploads, and clears it on completion or interruption. Adds light and dark UI previews and removes the debug connection-state dump.
- Persists the note pagination cursor together with refreshed notes so a failed refresh cannot advance past unstored results.

## Deployment

Deploy the offer-service changes before releasing the mobile app, which uses the new repost-recipient deletion endpoint.

## Summary by CodeRabbit

- **Improvements**
  - Offers, notes, and connections now refresh only affected recipients.
  - Failed updates retain retryable work for later attempts.
  - Existing connection information is preserved when fresh data is unavailable.
  - Background updates avoid unnecessary full connection refreshes.
  - Improved handling of social, club, and repost connection changes.
  - Added support for replacing and removing private recipients from reposted notes.
  - Added a visible status while offers are encrypted for new connections.
  - Updated the debug screen label to “Refresh offers and connections.”
  - Storage flushes now report whether persistence succeeded or failed.

### DIFF
--- a/apps/mobile/src/components/DebugScreen/index.tsx
+++ b/apps/mobile/src/components/DebugScreen/index.tsx
@@ -82,9 +82,7 @@ import {upsertInboxOnBeAndLocallyActionAtom} from '../../state/chat/hooks/useCre
 import {clubsToKeyHolderAtom} from '../../state/clubs/atom/clubsToKeyHolderV2Atom'
 import {clubsWithMembersAtom} from '../../state/clubs/atom/clubsWithMembersAtom'
 import {type ClubWithMembers} from '../../state/clubs/domain'
-import connectionStateAtom, {
-  syncConnectionsActionAtom,
-} from '../../state/connections/atom/connectionStateAtom'
+import connectionStateAtom from '../../state/connections/atom/connectionStateAtom'
 import offerToConnectionsAtom, {
   deleteOrphanRecordsActionAtom,
   updateAndReencryptAllOffersConnectionsActionAtom,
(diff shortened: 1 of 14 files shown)


SamTremko:
@greptile review

SamTremko:
@greptile review

kaladivo (approved):

